PR #3, opened 2022-02-28
wc_sessionPropose method documentation

dukevantreta:
Adds reference documentation for the `wc_sessionPropose` request and response JSON payloads.

Check if all parameters are correct, if there are any optional fields missing or not labelled.

Add comments if there are still any unresolved questions.

### DIFF
--- a/wc-rpc-methods.md
+++ b/wc-rpc-methods.md
@@ -19,16 +19,19 @@ error: {
 
 ### wc_sessionPropose
 Used to propose a session through topic A. Requires a success response with associated data. 
--   Success response is equivalent to session approval.
--   Error response is equivalent to session rejection.
--   This method *might* require a special timeout, because it needs end-user interaction to respond.
+- Success response is equivalent to session approval.
+- Error response is equivalent to session rejection.
+- This method *might* require a special timeout, because it needs end-user interaction to respond.
+- Proposer must use the relay parameter selected and sent by the responder, if different than the proposed one.
 ```jsonc
 // wc_sessionPropose params
 {
-    "relay": {
-        "protocol": string,
-        "data": string // Optional
-    },
+    "relay": [
+        {
+            "protocol": string,
+            "data": string // Optional
+        },
+    ],
     "blockchainProposed": {
         "auth": string, // Optional
         "chains": [string] // CAIP-2 chain IDs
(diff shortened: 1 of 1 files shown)


wc-rpc-methods.md
```diff
@@ -14,10 +15,60 @@ error: {
...
+        }
+    }
+}
+```
+
+```jsonc
+// Success result
+{
+    "relay": {
+        "protocol": string,
```
dukevantreta, line 64: Why does the responder send a `relay` field? What happens if the value is not the same as the proposal?

wc-rpc-methods.md
```diff
@@ -14,10 +15,60 @@ error: {
...
+            "icons": [string]
+        }
+    }
+}
+```
+
+```jsonc
+// Success result
+{
+    "relay": {
```
dukevantreta, line 63: Is the response's `relay` field optional or required?

wc-rpc-methods.md
```diff
@@ -14,10 +15,60 @@ error: {
...
+            "types": [string]
+        },
+        "controller": {
+            "publicKey": string
+        }
+    },
+    "proposer": {
+        "publicKey": string,
+        "controller": bool,
+        "metadata": {
```
dukevantreta, line 50: Are all `metadata` fields required? Is any of them optional?

wc-rpc-methods.md
```diff
@@ -14,10 +15,60 @@ error: {
...
+    "permissions": {
+        "jsonrpc": {
+            "methods": [string]
+        },
+        "notifications": {
+            "types": [string]
+        },
+        "controller": {
+            "publicKey": string
+        }
```
jakubuid, line 45: Should controller be nested in permissions?

pedrouid: Agreed, controller should be removed here

wc-rpc-methods.md
pedrouid: Yes, all 4 fields are required

wc-rpc-methods.md
pedrouid: Required

wc-rpc-methods.md
pedrouid: Proposer would use the new relay instead

We should make the relay an array on proposal so that responder can choose

dukevantreta: Changed the payload to use an array and added a bullet point to describe the selection.

Can the responder send a relay value totally different from the options proposed?

wc-rpc-methods.md
dukevantreta: Removed.

PR #5, opened 2022-03-01
wc_pairingDelete & wc_sessionDelete

dukevantreta:
Adds descriptions for `wc_pairingDelete` and `wc_sessionDelete`.

### DIFF
--- a/wc-rpc-methods.md
+++ b/wc-rpc-methods.md
@@ -46,7 +46,7 @@ Used to inform the peer to close and delete a pairing. All associated sessions o
 ### wc_sessionExtend
 
 ### wc_sessionDelete
-Used to inform the peer to close and delete a session.
+Used to inform the peer to close and delete a session. The reason field should be a human-readable message defined by the SDK consumer to be shown on the peer's side.
 ```jsonc
 // wc_sessionDelete params
 {


dukevantreta:
How should the `reason` parameter be used, from a UX point of view? What's the recommended practice, is it a human-readable string to display on the peer's side?

wc-rpc-methods.md
```diff
@@ -22,6 +22,13 @@ error: {
 ### wc_pairingExtend
 
 ### wc_pairingDelete
+Used to inform the peer to close and delete a pairing. All associated sessions of the given pairing must also be deleted.
+```jsonc
+// wc_pairingDelete params
+{
+    "reason": string
+}
```
jakubuid:
```suggestion
// wc_pairingDelete params
{
    "reason": string
    "code": int
}
```
Why we removed code?

wc-rpc-methods.md
```diff
@@ -38,6 +45,13 @@ error: {
 ### wc_sessionExtend
 
 ### wc_sessionDelete
+Used to inform the peer to close and delete a session.
+```jsonc
+// wc_sessionDelete params
+{
+    "reason": string
+}
```
jakubuid, line 55: Same

jakubuid:
> How should the `reason` parameter be used, from a UX point of view? What's the recommended practice, is it a human-readable string to display on the peer's side?

Whenever the session delete event is received by the peer, we return it to the skd's customer and display this message to the user.

wc-rpc-methods.md
dukevantreta: It didn't exist on the example payloads here: https://docs.walletconnect.com/2.0/protocol/client-communication
I'll be adding it, anyways

wc-rpc-methods.md
dukevantreta: Done

PR #6, opened 2022-03-01
wc_pairingExtend & wc_sessionExtend

dukevantreta:
Adds a description for `wc_pairingExtend` and `wc_sessionExtend`.

### DIFF
--- a/wc-rpc-methods.md
+++ b/wc-rpc-methods.md
@@ -19,16 +19,6 @@ error: {
 
 ### wc_sessionPropose
 
-### wc_pairingExtend
-Used to extend the lifetime of a pairing.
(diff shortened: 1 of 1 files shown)


wc-rpc-methods.md
```diff
@@ -20,6 +20,14 @@ error: {
 ### wc_sessionPropose
 
 ### wc_pairingExtend
```
pedrouid: We have decided to remove wc_pairingExtend all together and just create rules for pairings to extend expiry based on session approvals and rejections instead

dukevantreta: Removed wc_pairingExtend

PR #7, opened 2022-03-01
wc_sessionUpgrade method documentation

dukevantreta:
Adds payload documentation for `wc_sessionUpgrade`. Seems be the same as before, nothing new to see here.

### DIFF
--- a/wc-rpc-methods.md
+++ b/wc-rpc-methods.md
@@ -34,6 +34,23 @@ error: {
 ### wc_sessionUpdate
 
 ### wc_sessionUpgrade
+Used to upgrade the permissions of an active session. New permissions are merged over current permissions.
+```jsonc
+// wc_sessionUpgrade params
+{
+    "permissions": {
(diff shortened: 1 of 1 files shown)


wc-rpc-methods.md
```diff
@@ -34,6 +34,23 @@ error: {
...
 
 ### wc_sessionUpgrade
+Used to upgrade the permissions of an active session. New permissions are merged over current permissions.
+```jsonc
+// wc_sessionUpgrade params
+{
+    "permissions": {
+        "blockchain": {
+            "chains": [string] // CAIP-2 chain IDs
+        },
```
jakubuid, line 44: We agreed that we remove chains from the `wc_sessionUpgrade`.  Updating chains is moved to session update method.

Screenshot 2022-03-02 at 07 42 55

dukevantreta:
wc_sessionUpgrade method was changed to wc_sessionUpdateMethods

PR #8, opened 2022-03-01
wc_sessionRequest method

dukevantreta:
Adds the `wc_sessionRequest` method documentation.

### DIFF
--- a/wc-rpc-methods.md
+++ b/wc-rpc-methods.md
@@ -30,6 +30,17 @@ error: {
 ### wc_sessionSettle
 
 ### wc_sessionRequest
+Sends a CAIP-27 request to the peer client. The client should immediately reject the request and respond an error if the target session permissions doesn't include the requested method or chain ID.
+```jsonc
+// wc_sessionRequest params
+{
+    "request": {
+        "method": string,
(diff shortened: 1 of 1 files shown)


dukevantreta:
On google docs, the success acknowledgement response is a common boolean like all other methods. If this is the intended response for this method, we need a response method (`wc_sessionResponse`) to respond with the outcome of the request. For example, an approval or rejection of a transaction.

jakubuid:
> On google docs, the success acknowledgement response is a common boolean like all other methods. If this is the intended response for this method, we need a response method (`wc_sessionResponse`) to respond with the outcome of the request. For example, an approval or rejection of a transaction.

Yea good observation. I believe it's by accident. I vote to keep result for request the response. Imo there is no need for `wc_sessionResponse`.

llbartekll:
`wc_sessionResponse` doesn't make sense 🤣

pedrouid:
The reason you wouldn't need wc_sessionResponse is because the JSON-RPC standard already has binding between a request and a response

This means that a wc_sessionRequest request will always expect a corresponding response with the same JSON-RPC id

PR #10, opened 2022-03-03
wc_sessionSettle

dukevantreta:
Payload for `wc_sessionSettle`

### DIFF
--- a/wc-rpc-methods.md
+++ b/wc-rpc-methods.md
@@ -36,12 +36,9 @@ Used to settle a session over topic B.
         "protocol": string,
         "data": string // Optional
     },
-    "state": ???,
-    "blockchainSettled": {
-        "auth": string, // Optional
+    "blockchain": {
         "chains": [string], // CAIP-2 chain IDs
         "accounts": [string], // CAIP-10 account IDs
-        "signatures": ???
     },
     "permissions": {
         "jsonrpc": {


wc-rpc-methods.md
```diff
@@ -28,6 +28,42 @@ error: {
...
+    "relay": {
+        "protocol": string,
+        "data": string // Optional
+    },
+    "state": ???,
+    "blockchainSettled": {
+        "auth": string, // Optional
+        "chains": [string], // CAIP-2 chain IDs
+        "accounts": [string], // CAIP-10 account IDs
+        "signatures": ???
```
pedrouid: Let's remove "auth" and "signatures" for now... There is already so much to handle with this refactor... better to do it afterwards

Also can we rename "blockchainSettled" to "blockchain"

wc-rpc-methods.md
```diff
@@ -28,6 +28,42 @@ error: {
...
 ### wc_sessionSettle
+Used to settle a session over topic B.
+```jsonc
+// wc_sessionSettle params
+{
+    "relay": {
+        "protocol": string,
+        "data": string // Optional
+    },
+    "state": ???,
```
jakubuid: What about this one? We expose accounts by "blockchain" field so "state" is not needed here, imo.

dukevantreta: I agree

PR #28, opened 2022-04-11
add data-structures.md

pedrouid:


data-structures.md
```diff
@@ -0,0 +1,174 @@
...
+
+```jsonc
+// Approval
+{
+	"relay": {
+		"protocol": string,
+		"data": string // Optional
+	},
+	"responder": {
+		"publicKey": string
```
jakubuid, line 128: We should include responder metadata as well.
It's needed to update pairing responder metadata, once dApp receives session approval.

PR #29, opened 2022-04-11
add client-api.md

pedrouid:


xzilja:
Based on granularity of rpc methods, I'd propose to 

1. Delete ping and change it to `pingSession` and `pingPairing`
1. Delete disconnect and change it to `deleteSession` and `deletePairing`

xzilja:
We currently have few additional methods for events that are not specified here `once` and `off`. Might be worth including for better visibility. I also propose to get rid of `removeListener` as it is same as `off`, what do you think?

PR #34, opened 2022-04-14
Add "relayProtocol" optional param to approve()

pedrouid:


client-api.md
```diff
@@ -28,6 +28,7 @@ abstract class Client {
     accounts: string[];
     methods: string[];
     events: string[];
+    relayProtocol?: string;
```
xzilja, line 31: Perhaps just `relay` so it is concise with https://github.com/WalletConnect/sdk-team/blob/main/wc-rpc-methods.md#wc_sessionsettle ?

PR #47, opened 2022-05-10
Add session proposal overview

Elyniss:


pedrouid:
Needs updating after #48 is merged

pedrouid:
This is awesome @Elyniss 👌

Slight nitpick here kinda… Could we make sure that we are [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) compliant for this test cases 

This mean to always use the following terms MUST, MUST NOT, SHOULD, SHOULD NOT and MAY

Just saying this to avoid terms like “mustn’t” for clarity and also capitalized for quicker readability

Elyniss:
Good idea! I will update it right away

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+    },
+    "cosmos":{
+        "chains": ["cosmos:cosmoshub-4"],
+        "methods": ["cosmos_signDirect"],
+        "events": ["someCosmosEvent"]
+    }
+}
+```
+
+Wallet then validates if recieved proposal namespaces are valid. If they are not valid then session cannot be established and is rejected. If they are valid then Wallet decides if it should approve the proposal.
```
jakubuid: Could you elaborate on what you mean by saying: "Wallet then validates if received proposal namespaces are valid"?

jakubuid: Shouldn't wallet perform validation only when approving a session proposal?

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+```json
+{
+    "someSpecialKey":{
+        "chains": ["eip155:1", "cosmos:cosmoshub-4", "bip122:12a765e31ffd4059bada1e25190f6e98", "cosmos:Binance-Chain-Tigris"],
+        "methods": ["personalSign"],
+        "events": ["chainChanged"]
+    }
+}
+```
+Is valid?: Yes/No?
```
jakubuid: **1.4.**
It depends on what you mean by `"someSpecialKey"`. Imo for eip155 namespace we should keep only eip155 chains, for cosmos only cosmos chains etc. 
So at least for now (beta100) it shouldn't be allowed.

Elyniss:
Test cases SHOULD now be RFC-2199 compliant

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+        "chains": ["eip155:1", "eip155:10"],
+        "methods": ["personalSign"],
+        "events": []
+    },
+    "someSpecialKey":{
+        "chains": ["eip155:1"],
+        "methods": ["eth_getAccounts"],
+        "events": []
+    }
+}
```
jakubuid, line 152: **1.5.**
Again, it depends on what you mean by `"someSpecialKey"`. 
That's why we introduced extensions, to have a special method among one namespace for given chain.
Hence, imo `1.5` is invalid.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+```
+Is valid?: No
+
+Note: Even though, the first proposal namespace is valid the second being invalid makes whole proposal invalid
+
+---
+
+
+Czy dodawać walidacje tego co jest wewntarz przestrzeni?? eip155 musi jedynie zawierac eip155:42 itp?
```
jakubuid: Should be removed :)

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+        "methods": ["cosmos_signDirect"],
+        "events": ["someCosmosEvent"]
+    }
+}
+```
+Received session namespaces:
+```json
+{
+    "cosmos": {
+        "chains": [],
```
jakubuid:
```suggestion
        "accounts": [],
```

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+
+### 2.7. TODO: Multiple accounts from different origin in one namespace
+Requested proposal namespaces:
+```json
+{
+    "someSpecialKey": {
+        "chains": ["eip155:1", "cosmos:cosmoshub-4"],
+        "methods": ["personalSign"],
+        "events": ["accountsChanged"]
+    }
```
jakubuid, line 442: **2.7.**
Same as above. But I might missing some insights where mixing chains is possible.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+```json
+{
+    "eip155": {
+        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:42:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+        "methods": ["eth_sign"],
+        "events": ["accountsChanged"]
+    }
+}
+```
+Is valid?: Yes/No
```
jakubuid: **2.8.**
If we assume that requested proposal namespaces are min requirements, it is valid then.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+```json
+{
+    "someSpecialKey": {
+        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb","eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "cosmos:cosmoshub-4:cosmos1t2uflqwqe0fsj0shcfkrvpukewcw40yjj6hdc0"],
+        "methods": ["eth_sign", "cosmos_signDirect"],
+        "events": ["accountsChanged", "someCosmosEvent"]
+    }
+}
+```
+Is valid?: Yes/No
```
jakubuid: **2.9.**
Imo it's invalid. We should keep the same structure of requested and received namespaces.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
```
jakubuid: **2.10.**
Imo we should avoid such merging, for me it creates confusion.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+        "events": ["accountsChanged"]
+    },
+    "eip155ButBetter":{
+        "accounts": ["eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+        "methods": ["eth_sign"],
+        "events": ["accountsChanged"]
+    }
+}
+```
+Is valid?: Yes/No
```
jakubuid: **2.11.**
Same, I think we should avoid decomposition of the same namespace.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,559 @@
...
+```json
+{
+    "eip155": {
+        "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+        "methods": ["eth_sign", "personalSign"],
+        "events": ["accountsChanged"],
+    }
+}
+```
+Is valid?: Yes/No
```
jakubuid: **2.12.**
I think that the reason of having extensions is to actually highlight the difference between chains. Once merging them we reduce the importance of extensions.

auth/session-proposal-overview.md
Elyniss: > Could you elaborate on what you mean by saying: "Wallet then validates if received proposal namespaces are valid"?

Before any approval, I reckon we need to verify if proposal namespaces are valid. If the wallet receives malformed proposal namespaces its unable to approve session and respond with valid session namespaces.

jakubuid: Ya, make sense. Therefore, wallet should respond internally with proper error code to let the Dapp know that it sent invalid namespaces.

Elyniss: Exactly my initial thought. I will rephrase the docs so it is clear.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,556 @@
...
+    },
+    "cosmos":{
+        "chains": ["cosmos:cosmoshub-4"],
+        "methods": ["cosmos_signDirect"],
+        "events": ["someCosmosEvent"]
+    }
+}
+```
+
+Wallet then validates if recieved proposal namespaces are valid. If they are not valid then session cannot be established and wallet rejects it with some special reason and code that tells the Dapp the proposal namespaces are invalid (TODO: Define reason and code) . If they are valid then Wallet decides if it should approve the proposal.
```
Elyniss: @jakobuid Does this look better? If yes, can you help me with defining reason and code? Do we have such code defined somewhere?

jakubuid: Ya, it looks better now.
In terms of error code, I'd say: `case .invalidSessionProposeRequest: return 1006`

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,556 @@
...
+    },
+    "cosmos":{
+        "chains": ["cosmos:cosmoshub-4"],
+        "methods": ["cosmos_signDirect"],
+        "events": ["someCosmosEvent"]
+    }
+}
+```
+
+Wallet then validates if recieved proposal namespaces are valid. If they are not valid then session cannot be established and wallet rejects it with `1006` code that tells the Dapp the proposal namespaces are invalid. If they are valid then Wallet decides if it should approve the proposal.
```
Elyniss: @jakobuid What do you think about this? I've also added `case .invalidSessionProposeRequest: return 1006` to `auth/error-codes.md`

jakubuid: Ya, I like this.
Just being curious what other @WalletConnect/sdk-team members think about it.

Previously, we've been showing the session proposal to the user without any validation but now we could change it to show only those which are valid.
If the namespace in the session proposal is invalid, we could send the internal error code to a Dapp (`case .invalidSessionProposeRequest: return 1006`) and do not show session proposal to the user.

Obviously, this doesn't mean that we should omit namespaces validation when approving session. It should be done regardless.

What do you think guys?

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,556 @@
+# Session proposal overview:
+Let's say a dapp wants access to Ethereum Mainnet, Polygon, Cosmos Mainnet.
+It issues required methods and events for specific blockchains in form of  proposal namespaces. For example Ethereum Mainnet and Polygon it requires: `eth_sign` method and `accountsChanged` event, and for Cosmos Mainnet it requires: `cosmos_signDirect` method and `someCosmosEvent` event. Let's say that Polygon also has special method `personalSign` and event `chainChanged` not available for Ethereum Mainnet. Dapp can request those only for Polygon via `extensions` field.
```
dukevantreta:
```suggestion
It specifies the proposed execution environment for each blockchain in the form of namespaces. For example, for Ethereum Mainnet and Polygon, it requires: `eth_sign` method and `accountsChanged` event, and, for Cosmos Mainnet, it requires: `cosmos_signDirect` method and `someCosmosEvent` event. Let's say that Polygon also has a special method `personalSign` and an event `chainChanged` not available in Ethereum Mainnet. The Dapp can require chain-exclusive parameters via the `extensions` field. 
```

auth/session-proposal-overview.md
dukevantreta:
```suggestion
The Wallet then validates if the received proposal namespaces are valid. If they are not valid, then the session cannot be established and the Wallet rejects it with a `1006` code that tells the Dapp that the proposal namespaces are invalid. If they are valid, then the Wallet if free to decide whether to approve the proposal, or reject it.
```

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,556 @@
...
+        "chains": ["cosmos:cosmoshub-4"],
+        "methods": ["cosmos_signDirect"],
+        "events": ["someCosmosEvent"]
+    }
+}
+```
+
+Wallet then validates if recieved proposal namespaces are valid. If they are not valid then session cannot be established and wallet rejects it with `1006` code that tells the Dapp the proposal namespaces are invalid. If they are valid then Wallet decides if it should approve the proposal.
+
+If Wallet doesn't approve then session is rejected. Otherwise, Wallet responds with slighly different structure than proposal namespaces called session namespaces. Instead of list of `chains` it has list of `accounts` that approves given methods and events. If wallet approves session proposal it needs to accept all methods and events in every proposal namespace. If Wallet wants it can give permissions for more methods and events than requested, but never less.
```
dukevantreta:
```suggestion
If the Wallet (or the user) does NOT approve the session, then it is rejected. Otherwise, the Wallet responds with a slightly different namespace schema: session namespaces. Instead of having a list of `chains`, it has list of `accounts` compatible with the given methods and events. If the Wallet approves a session proposal, it needs to accept all methods and events of all proposal namespaces. If needed, the Wallet can add permissions for more methods and events than the ones requested, but never less.
```

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,556 @@
...
+    },
+    "cosmos":{
+        "accounts": ["cosmos:cosmoshub-4:cosmos1t2uflqwqe0fsj0shcfkrvpukewcw40yjj6hdc0"],
+        "methods": ["cosmos_signDirect", "personal_sign"],
+        "events": ["someCosmosEvent", "proofFinalized"]
+    }
+}
+```
+
+Dapp then validates if recieved session namespaces are valid for requested proposal namespaces. If they are valid then session is established. If they are not valid then session cannot be established.
```
dukevantreta:
```suggestion
The Dapp then validates if the received session namespaces comply with the requested proposal namespaces. If so, the session is established / settled. If not, the session is not established and MUST delete all related cached data.
```

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,556 @@
...
+}
+```
+Is valid?: No
+
+Note: `0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb` is not CAIP-10 compliant.
+`eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb` is CAIP-10 compliant
+
+---
+
+### 2.3. Session namespaces MUST accept all methods and events
```
dukevantreta: If we mean to use this as a guide for test cases, I believe it's useful to split this point into 2, so that methods and events validation can be tested separately and in isolation

Elyniss: Good call! I will add test case **2.13.** for events and change this one to only test methods

auth/session-proposal-overview.md
xzilja: We have it set up in js like this

1. If at any point users pass invalid params (so session, pairing etc... objects) client will throw an error. This way it is up to dapp devs to resolve the issue. Nothing gets sent to relay / persisted.

2. Same happens when incoming request payload is invalid, we just don't emit an event and instead log error (so it is silent on wallet side).

auth/session-proposal-overview.md
Elyniss: I see your point. But the namespace is still logically valid. Session namespaces gave access to all requested methods/events/chains. The Dapp asked for `personalSign` only for `eip155:137` however, the wallet gave `personalSign` method permission to `eip155:1` as well.

I don't think we reduce the importance of extensions because this approach should not be encouraged. But making this test case invalid goes against our idea that the wallet can approve what he wants but MUST at least approve proposal namespaces. 
 
Thinking about it gave birth to 2.14, 2.15, 2.16. Be sure to check those out as well.

auth/session-proposal-overview.md
Elyniss: To iterate over your comment. Should we add rule saying: `All chains in the namespace MUST contain the namespace prefix`? 

Valid example:
```json
{
    "sthWeird": {
        "chains": ["sthWeird:chainReference"],
        "methods": [],
        "events": []
    }
}
```
Invalid example:

```json
{
    "sthWeird": {
        "chains": ["sthElse:chainReference"],
        "methods": [],
        "events": []
    }
}
```

jakubuid: Yes, I think we should add such rule, because otherwise we could end up having chains not related to actual namespace.

auth/session-proposal-overview.md
jakubuid: Ok, after your explanation I can confirm that's valid. You're right saying that wallet sending above session namespaces conforms the requested namespace, and because the fact that it's controller it could add `personalSign` to `eip155:1` as well.

auth/session-proposal-overview.md
jakubuid: If there's malicious Dapp, there is no check of namespaces before sending a proposal to a wallet. Hence, imo we should add such validation internally on wallet side, when wallet receives session proposal.

If namespaces in session proposal are invalid, wallet sends JsonRpcError to the Dapp OR log error silently as you suggested.

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,694 @@
...
+            {
+                "chains" : ["eip155:10"],
+            }
+        ]
+    }
+}
+```
+Is valid?: Yes
+
+Note: First extension is missing `events` field but remains valid. Second extensions is missing `methods` field but remains valid. Third extensions only has `chains` field but due to `method` and `events` field being optional it remains valid, even though it has no logical value.
```
Talhaali00: So we decided that methods and events cannot be optional and the default state is an empty array

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,694 @@
...
+    }
+}
+```
+Is valid?: No
+
+Note: Even though, the first proposal namespace is valid the second being invalid makes whole proposal invalid
+
+---
+
+## Non-controller side validation of incoming proposal namespaces (Dapp)
```
Talhaali00: This is the first instance of Non-Controller used in the doc. Should we replace all instances of wallet/dapp with controller/non-controller?

auth/session-proposal-overview.md
Talhaali00: I agree, the key must exist in the chain and account info

auth/session-proposal-overview.md
Talhaali00: Maybe we make it a requirement that the Session Namespaces must contain at least the same keys as the Proposal Namespaces

auth/session-proposal-overview.md
Talhaali00: But we have an issue with the structure. The Proposal has an extension but the Session is missing the extension. I believe this would make Session Namespace invalid. I think that adding to the general methods is fine but if there are extensions requested by the Dapp, the Wallet must also respond with extensions

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,694 @@
...
+            {
+                "accounts": ["eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+                "methods": ["eth_sign", "personalSign"],
+                "events": ["accountsChanged"]
+            }
+        ]
+    }
+}
+```
+Is valid?: Yes/No
```
Talhaali00: I think this is valid. It follows the same logic as well the general methods array

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,694 @@
...
+            {
+                "accounts": ["eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:42:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+                "methods": ["personalSign"],
+                "events": ["accountsChainged"]
+            }
+        ]
+    }
+}
+```
+Is valid?: Yes/No
```
Talhaali00: I think this is also valid. Wallet is giving more permissions than requested which is allowed

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,694 @@
...
+            {
+                "accounts": ["eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"],
+                "methods": ["personalSign"],
+                "events": ["accountsChainged"]
+            }
+        ]
+    }
+}
+```
+Is valid?: Yes/No
```
Talhaali00: I think this is also allowed. Wallet can add anything as long as it meets the minimum requirements of the Proposal Namespace

jakubuid:
> Should we also specify what the error code and message will be when validation fails?

This is already specified, see invalidUpdateRequest: https://github.com/WalletConnect/sdk-team/blob/main/auth/error-codes.md#invalid

auth/session-proposal-overview.md
llbartekll: imo definitely not. from caip-2:
```
chain_id:    namespace + ":" + reference
namespace:   [-a-z0-9]{3,8}
reference:   [-a-zA-Z0-9]{1,32}
```
I understand it like `someSpecialKey` must be a `namespace` from caip-2 chain id
and Namespace(our new structure) chains are caip-2s with "parent" chain_id namespace
so chain_id/namespace must be equal in all chains

auth/session-proposal-overview.md
llbartekll: agree, invalid

auth/session-proposal-overview.md
llbartekll: agree

auth/session-proposal-overview.md
llbartekll: that `someSpecialKey` is so confusing xd

I would remove all the test cases with `someSpecialKey` or leave at most once to show that is forbidden to mix chains in namespaces

auth/session-proposal-overview.md
llbartekll: would remove this test case as well as it has the same exclusion

auth/session-proposal-overview.md
llbartekll: would remove test case

auth/session-proposal-overview.md
llbartekll: scope is bigger than required so IMO **valid**

auth/session-proposal-overview.md
llbartekll: imo valid

auth/session-proposal-overview.md
llbartekll: `accountsChainged` is unnecessary in the extension so it extends nothing, I would say it is incorrect but not necessarily a reason to throw an error.

auth/session-proposal-overview.md
llbartekll: To me it does not make sense. technically it extends a chain `eip155:42` that is not in a basic scope. I do not think we need such accurate validation of Namespaces and it does not need to throw but IMO it is not correct or at least not something anyone want to use.

auth/session-proposal-overview.md
Elyniss: I marked it as valid.

auth/session-proposal-overview.md
Elyniss: Thanks for noticing! Remade the test case

auth/session-proposal-overview.md
Elyniss: I did not see the need to update the overview.

auth/session-proposal-overview.md
Elyniss: Marked it valid.

auth/session-proposal-overview.md
Elyniss: Marked it valid.

Elyniss:
> We should probably also add validation that a key cannot be empty

I've added namespace key validation as specified [here](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)

Elyniss:
> I think there is too many test cases with the same exclusion reason(someSpecialKey) and some that are a little unrealistic and not harmful.

I've removed/restructured unrealistic ones.

Elyniss:
> This is already specified, see invalidUpdateRequest: https://github.com/WalletConnect/sdk-team/blob/main/auth/error-codes.md#invalid

I have also added `case .invalidSessionProposeResponse: return 1007`. Which is sent by non-controller when Session Namespaces are invalid.

jakubuid:
> I have also added `case .invalidSessionProposeResponse: return 1007`. Which is sent by non-controller when Session Namespaces are invalid.

@Elyniss could you elaborate on that? when non-controller should send such error?

Elyniss:
> > I have also added `case .invalidSessionProposeResponse: return 1007`. Which is sent by non-controller when Session Namespaces are invalid.
> 
> @Elyniss could you elaborate on that? when non-controller should send such error?

So, during the review, we come up with an idea to add an error when the Controller receives an invalid/malformed Proposal Namespaces in request. We decided that Controller should tell the Non-controller that it send the wrong request by responding with `1006` error.

I reckon there is a similar situation when the Non-controller receives Session Namespaces in response. I propose we add another code `1007` which the Non-controller should send to the Controller after it sends invalid/malformed Session Namespaces as a response to the Session Proposal. For example, Controller did not agree on one method in Proposal Namespaces. Non-controller should have some way of telling him that something is wrong. 

@jakobuid Is this clear enough?

jakubuid:
> > > I have also added `case .invalidSessionProposeResponse: return 1007`. Which is sent by non-controller when Session Namespaces are invalid.
> > 
> > 
> > @Elyniss could you elaborate on that? when non-controller should send such error?
> 
> So, during the review, we come up with an idea to add an error when the Controller receives an invalid/malformed Proposal Namespaces in request. We decided that Controller should tell the Non-controller that it send the wrong request by responding with `1006` error.
> 
> I reckon there is a similar situation when the Non-controller receives Session Namespaces in response. I propose we add another code `1007` which the Non-controller should send to the Controller after it sends invalid/malformed Session Namespaces as a response to the Session Proposal. For example, Controller did not agree on one method in Proposal Namespaces. Non-controller should have some way of telling him that something is wrong.
> 
> @jakobuid Is this clear enough?

If I understand you reply correctly, this approach is wrong. Here is why.
When non-controller sends a request it awaits to receive response from controller. Upon that, it's the end of the request response cycle. Any request can have only one response. You can't send a request, receive response and send response, because there is no new request that awaits for response.

Is it clear enough?

Elyniss:
Oh, makes sense. Thanks!

llbartekll:
maybe we could define a rule:
`extension is an Array of namespaces`? so we can limit test cases and errors.
 to me they are to verbose.

@WalletConnect/sdk-team

Elyniss:
> maybe we could define a rule: `extension is an Array of namespaces`? so we can limit test cases and errors. to me they are to verbose.
> 
> @WalletConnect/sdk-team

I'm against it. Since namespaces contain extensions, defining extensions as an array of namespaces would mean nested extension

llbartekll:
> > maybe we could define a rule: `extension is an Array of namespaces`? so we can limit test cases and errors. to me they are to verbose.
> > @WalletConnect/sdk-team
> 
> I'm against it. Since namespaces contain extensions, defining extensions as an array of namespaces would mean nested extension

So this can be a statement in that rule or maybe it make sense to have nested extensions

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,696 @@
...
+        "events": ["someCosmosEvent", "proofFinalized"]
+    }
+}
+```
+
+The Dapp then validates if the received Session Namespaces comply with the requested Proposal Namespaces. If so, the session is established / settled. If not, the session is not established and MUST delete all related cached data.
+
+# Validation test cases
+
+## Controller side validation of incoming Proposal Namespaces (Wallet)
```
Talhaali00, line 62: @Elyniss Instead of making this the logic done on the wallet side, it would be better to make the Dapp does this validation before even sending the namespace to the wallet

Talhaali00: client should validate peer's namespaces to protect against malicious peers

auth/session-proposal-overview.md
```diff
@@ -0,0 +1,696 @@
...
+```
+Is valid?: No
+
+Note: Proposal namespace extension doesn't have any chains, hence it's invalid
+
+Throw Message: `Chains must not be empty`
+
+---
+
+### 1.6. Extension methods field is REQUIRED
```
dukevantreta, line 162: By "required", do you mean only non-null or also non-empty? An array field can be required and still be empty. (Same question for events field, `1.7`)

dukevantreta:
I feel missing a simple rule or spec regarding empty strings for methods and events. We should also invalidate these, no?

llbartekll:
> > > maybe we could define a rule: `extension is an Array of namespaces`? so we can limit test cases and errors. to me they are to verbose.
> > > @WalletConnect/sdk-team
> > 
> > 
> > I'm against it. Since namespaces contain extensions, defining extensions as an array of namespaces would mean nested extension
> 
> So this can be a statement in that rule or maybe it make sense to have nested extensions

@Elyniss Actually I can recall, someone already suggested nested extensions and we agreed on not having them. So an extension is an array of non-extendable namespaces. Anyway just a suggestion to reduce test cases, do what you want.

auth/session-proposal-overview.md
Elyniss: Events and methods can be empty. They must not be null or nonexistent.

Elyniss:
> I feel missing a simple rule or spec regarding empty strings for methods and events. We should also invalidate these, no?

@MisterVants Personally, I don't think that is necessary. But if we were to add them, should we also add some length restrictions as well?

Elyniss:
> > > > maybe we could define a rule: `extension is an Array of namespaces`? so we can limit test cases and errors. to me they are to verbose.
> > > > @WalletConnect/sdk-team
> > > 
> > > 
> > > I'm against it. Since namespaces contain extensions, defining extensions as an array of namespaces would mean nested extension
> > 
> > 
> > So this can be a statement in that rule or maybe it make sense to have nested extensions
> 
> @Elyniss Actually I can recall, someone already suggested nested extensions and we agreed on not having them. So an extension is an array of non-extendable namespaces. Anyway just a suggestion to reduce test cases, do what you want.

@llbartekll Given it some thought, I am starting to like it. I will add the rule and see how it looks

PR #53, opened 2022-05-30
update error codes

Elyniss:


sign/error-codes.md
```diff
@@ -29,5 +30,11 @@ case .invalidMethod: return 1001
...
 
 case .invalidUpdateRequest: return 1003
-case .invalidUpdateExtendRequest: return 1004
+case .invalidExtendRequest: return 1004
+```
+
+## REMAINING
+
+```sh
+case .userDisconnected: return 6000
```
jakubuid, line 41: Haven't we agreed to add more disconnection reasons as well?

sign/error-codes.md
```diff
@@ -29,5 +30,12 @@ case .invalidMethod: return 1001
...
+case .sessionSettlementFailed: return 6001
```
Elyniss: In Kotlin SDK there is a chance that the session cannot be settled due to a database error. Then return this error code.

Elyniss:
Complemented test cases with their error codes

TalhaAli00 (approved):

IljaDaderko (approved):

sign/error-codes.md
llbartekll: 6000 and 6001 are not much related so maybe for `sessionSettlementFailed` you should give 7000 or something else?
another thing is that `userDisconnected` is actually not an error but more like a reason code. it is included as a param in a request in contrast to all others that are included in error responses.

so should we call the doc `error-codes.md`?

Elyniss: I see what you mean with `userDisconnected`. I've updated the specs. What do you think of them now?

llbartekll: `protocol-codes.md` ?
now I like it more.

llbartekll (approved):

sign/error-codes.md
Elyniss: I would rather stick with `codes.md` rather than `protocol-codes.md`. IMO, sounds/reads better.

PR #58, opened 2022-05-30
Chat MVP

pedrouid:


chat/discoverability.md
```diff
@@ -0,0 +1,35 @@
...
+
+Thus A can query the ENS domain and resolve the mapping to the chat public key
+
+## Scenario 3 - B shares public key with A through the qrcode
+
+In this scenario, it is assumed that A does not have knowledge of B's blockchain account.
+
+We assume that A and B have other means of communication such as a face-to-face interaction or an online video call where they can exchange a qrcode to send a chat invite to each other.
+
+Thus A can scan the qrcode provided by B to retrieve blockchain account and public key
```
chadyj, line 35: A UX question: Will we just have one "scan QR code" functionality?

It would be preferable to avoid a confusing UX situation where a wallet has 3 different QR code scanning buttons for transactions, sign, and chat.

chat/README.md
```diff
@@ -2,4 +2,27 @@
...
+
+Chat functionality would be present in wallets where you could search or add a new contact from their wallet addres on-chain and start a one-to-one conversation with them.
+
+It would resemble direct messaging (DM) currently present in many social media applications but it would be end-to-end encrypted and available across different wallet providers.
+
+Additionally it could be recovered from a seed phrase to allow users to migrate wallet provider without loosing their chat.
+
+## Goals
+
+A user can register their blockchain account to be discoversable in the public-key registry.
```
chadyj: typo: discoverable

chat/client-api.md
```diff
@@ -0,0 +1,46 @@
...
+  }): Promise<void>;
+
+  // ---------- Events ----------------------------------------------- //
+
+  // subscribe to new chat invites received
+  public abstract on("new_invite", ({ id: number, openingMessage: string }) => {}): void;
+
+  // subscribe to new chat messages received
+  public abstract on("new_message", ({ topic: number, message: string }) => {}): void;
+
```
llbartekll, line 85: to add on_newThread ?

llbartekll: and on_leave

chat/invite-uri.md
```diff
@@ -0,0 +1,33 @@
...
+    parameters  = parameter *( "&" parameter )
+    parameter   = key "=" value
+    key         = STRING
+    value       = STRING
+
+## Parameters
+
+### Required
+
+- `account` (STRING) = account associated with public key
```
llbartekll, line 17: should it be really required?
IMO for qrcodes it is not necessary

chat/discoverability.md
pedrouid: That's correct. URI schemas can be machine readable to identify which action the user is scanning

So whether it's an address, transaction, connection or chat... they will be "routed" to the correct client

chat/client-api.md
pedrouid: Cool, I will add these

chat/invite-uri.md
pedrouid: The mapping will always relate to a blockchain account and a chat public key

So if you scan the qrcode for the public key then you don't really have a "label" for this new contact

I understand the privacy concerns but for MVP we should stick to CAIP-10 accounts

llbartekll: > The mapping will always relate to a blockchain account and a chat public key

that is true for registry

> So if you scan the qrcode for the public key then you don't really have a "label" for this new contact

that is also true but there could be an extra param like `name/label` or even person that scans could set the name by it's own like `my local egg dealer` or whatever

llbartekll: but sure we can assume it is required caip-10 for now

chat/README.md
```diff
@@ -2,4 +2,27 @@
 
 ## Description
 
-Chat API allows wallet users to message 1-on-1 with other wallet users in the WalletConnect network by using blockchain accounts as addresses to establish secure messaging between peers.
+Chat API allows wallet users to message 1-on-1 with other wallet users in the WalletConnect network by using blockchain accounts as addresses to establish end-to-end messaging between peers.
+
+## Context
+
+Chat functionality would be present in wallets where you could search or add a new contact from their wallet addres on-chain and start a one-to-one conversation with them.
```
bkrem:
```suggestion
Chat functionality would be present in wallets where you could search or add a new contact from their wallet address on-chain and start a one-to-one conversation with them.
```

chat/README.md
bkrem:
```suggestion
A user can register their blockchain account to be discoverable in the public-key registry.
```

chat/README.md
```diff
@@ -2,4 +2,27 @@
...
-Chat API allows wallet users to message 1-on-1 with other wallet users in the WalletConnect network by using blockchain accounts as addresses to establish secure messaging between peers.
+Chat API allows wallet users to message 1-on-1 with other wallet users in the WalletConnect network by using blockchain accounts as addresses to establish end-to-end messaging between peers.
+
+## Context
+
+Chat functionality would be present in wallets where you could search or add a new contact from their wallet address on-chain and start a one-to-one conversation with them.
+
+It would resemble direct messaging (DM) currently present in many social media applications but it would be end-to-end encrypted and available across different wallet providers.
+
+Additionally it could be recovered from a seed phrase to allow users to migrate wallet provider without loosing their chat.
```
bkrem:
```suggestion
Additionally it could be recovered from a seed phrase to allow users to migrate wallet provider without losing their chat.
```

chat/README.md
```diff
@@ -2,4 +2,27 @@
...
+
+Chat functionality would be present in wallets where you could search or add a new contact from their wallet address on-chain and start a one-to-one conversation with them.
+
+It would resemble direct messaging (DM) currently present in many social media applications but it would be end-to-end encrypted and available across different wallet providers.
+
+Additionally it could be recovered from a seed phrase to allow users to migrate wallet provider without losing their chat.
+
+## Goals
+
+A user can register their blockchain account to be discoverable in the public-key registry.
```
bkrem: I think there's another naming discussion to be had here at some point if we ever want to mention the chat's public key registry in public copy or documentation.

Otherwise this will probably be easily confused with the Cloud wallet/dapp registry since that's the existing association for people.

chat/README.md
```diff
@@ -2,4 +2,27 @@
...
+
+A user can invite another wallet user to chat with their blockchain account regardless if it's discoverable.
```
bkrem:
```suggestion
A user can invite another wallet user to chat with their blockchain account regardless of its discoverability.
```

chat/client-api.md
```diff
@@ -0,0 +1,52 @@
...
+  public abstract message(params: {
+    topic: string;
+    message: string;
+  }): Promise<void>;
+
+  // leaves a chat thread and stops receiving messages
+  public abstract leave(params: {
+    topic: string;
+  }): Promise<void>;
+
```
bkrem, line 53: Would it make sense to have something like `private abstract sync()` to synchronize the client with the remote source of truth? 

E.g.  on `init()` and other scenarios where it may be desirable to validate that local state is in sync with the remote (in case event listeners were paused on a long running tab etc)

chat/client-api.md
```diff
@@ -0,0 +1,52 @@
...
+  // queries the default registry with a blockchain account and returns the public key
+  public abstract resolve(params: {
+    account: string;
+  }): Promise<string>;
+
+  // sends a chat invite with opening message and returns an invite id
+  public abstract invite(params: {
+    publicKey: string;
+    openingMessage: string;
+  }): Promise<number>;
```
Elyniss: What would be an `invite id` use case?

chat/client-api.md
Elyniss: Good idea. We should discuss how to validate whether local and remote are in sync. Maybe a timestamp of the last message?

chat/client-api.md
Elyniss: How about `on_ping`?

arein:
It's a bit hard to review this without having a better high-level overview. I suggest that we go over the [chat project brief](https://docs.google.com/document/d/1brVsh0KoXgT88h1eIpNQ6g8N42xepY5WOdLhOgjfP90/edit#heading=h.c926nb8x5szd) to also come to a shared understanding of what we're building. Thoughts?

chat/client-api.md
llbartekll: I think sdk can respond on_ping without having an event

chat/README.md
llbartekll: Chat registry?

chat/chat-invite.md
```diff
@@ -0,0 +1,38 @@
...
+
+B derives symKey T using pubKey Y and privKey Z.
+
+Thread topic is derived as the hash of the symKey T.
+
+A receives acceptance which includes pubKey Z in the content.
+
+A derives symKey T using privKey Y and pubKey Z.
+
+A and B both subscribe to thread topic and sends messages with symKey T
```
llbartekll: I just realised that a protocol has changed from our alpha doc. 
@pedrouid can you share the motivation behind having two symKeys?

I imagined it as below:

```mermaid
sequenceDiagram
    participant Alice
    participant Bob
    participant Chat Registry
    Alice->>Chat Registry: registers her blockchain account and publicKey X
    Bob->>Chat Registry: resolves Alice's publicKey X by her blockchain account
    Bob->>Bob: Generates a keyPair Y.
    Bob->>Bob: Derives symKey T with DH using publicKey X and privateKey Y
    Bob->>Bob: Derives invite topic by hashing publicKey X
    Bob->>Bob: Encrypts openingMessage with symKey T
    Bob->>Bob: Derives thread topic by hashing symKey T and subscribes to it
    Bob->>Alice: Publishes invite on the invite topic.
    Alice->>Alice: Derives symKey T
    Alice->>Alice: Decrypts invite's opening message deriving DH symKey T.
    Alice->>Alice: Sends acceptance on invite topic encrypted with symKey T.
    Alice->>Alice: Subscribes on Thread topic
    Bob->>Alice: Sends a message to Alice
```

chat/client-api.md
llbartekll: @bkrem could you elaborate on that? what you mean by `remote source of truth` and ` local state` 

checking if massage has been acknowledged?

chat/invite-uri.md
pedrouid: I agree with your POV but if we focus on the major focus of this API is that it provides wallet-to-wallet chat on top of WalletConnect... we could add in the future arbitrary account/label support but it's broadening the scope too much

Also custom labels can and should be incentivized for Wallets to add on top of our protocol while we focus on a simple mapping for addresses to chat public keys

chat/client-api.md
pedrouid: Exactly. Similarly to how Sign API works

chat/README.md
pedrouid: it should be called keyserver. sorry my bad

chat/client-api.md
pedrouid: I think we need to also consider if this is a Chat responsibility or a Relay responsibility

chat/client-api.md
pedrouid: Perhaps tracking pending invites on the UI

chat/chat-invite.md
pedrouid: After our discussion, I have changed the topic I to be unencrypted thus only using the symKey for hiding opening message and acceptance

chat/client-api.md
bkrem: > @bkrem could you elaborate on that? what you mean by remote source of truth and local state

Basically I'm wondering if it makes sense to have an explicit `pull` approach in the client, besides the `on push` aspect provided via `on("chat_message",...)`, for a scenario similar to this:

1. Client A and B establish a chat thread and send messages
2. Client B goes offline/stops listening for `"chat_message"`.
3. Client A sends 3 more messages
4. Client B comes back online, but its local state is now 3 messages behind the expected shared state

To resolve step 4, do we:

1. push all unack'd messages to Client B when it reconnects (and as Pedro mentioned is this a relay responsibility?)
2. Let client B call something like `sync` to pull any messages it is missing whenever it deems it necessary to do so.

llbartekll: In my understanding 
3.a Client A sends 3 more messages
3.b Relay server ack receiving of 3 messages - message status [published](https://github.com/WalletConnect/walletconnect-specs/blob/chat-mvp/chat/message-status.md) 
4.a client B comes back and subscribes to thread topic and relay delivers all 3 "pending" at relay messages.
4.b. client B responds to all 3 messages(same jsonrpc id as request aka message) so A can recognise those messages as `delivered`

chat/chat-keyserver.md
```diff
@@ -0,0 +1,34 @@
...
+```jsonc
+{
+    “account”: string,
+    “publicKey”: string
+}
+```
+
+### Resolve
+
+Used for resolving record by account.
```
Elyniss:
```suggestion
Used for resolving record by CAIP10 account.
```

chat/rpc-methods.md
```diff
@@ -0,0 +1,95 @@
...
+
+### wc_chatMessage
+
+Used to send a message to its peer through topic T.
+
+- Success response is equivalent to message delivery receipt.
+- Error response is equivalent to message delivery failure.
+
+```jsonc
+// wc_chatInvite params
```
jakubuid:
```suggestion
// wc_chatMessage params
```

chat/client-api.md
bkrem: Gotcha, makes more sense to me now thanks, so simply resubscribing to the thread topic when becoming active again should resolve this `sync` scenario 👍

chat/rpc-methods.md
```diff
@@ -0,0 +1,95 @@
...
+    "type": string,
+    "data": string,
+  }
+}
+```
+
+```jsonc
+// Success result
+{
+  "receivedAt": Int64 // timestamp when received
```
Elyniss: Should this be used for ordering the messages in chat?

Elyniss:
Right now I don't see any message ordering mechanism. Am I missing something?

chat/rpc-methods.md
```diff
@@ -0,0 +1,91 @@
...
+  // empty
+}
+```
+
+### wc_chatLeave
+
+```jsonc
+// wc_chatLeave params
+{
+  "publicKey": string,
```
llbartekll: what is publicKey for? could it be an empty request?

chat/client-api.md
```diff
@@ -0,0 +1,61 @@
...
+  // queries the default keyserver with a blockchain account / returns the public key
+  public abstract resolve(params: {
+    account: string;
+  }): Promise<string>;
+
+  // sends a chat invite with opening message / returns an invite id
+  public abstract invite(params: {
+    publicKey: string;
+    message: string;
+  }): Promise<number>;
```
llbartekll: Now our `Invite` data model contains an `account` but api's invite method does not.
should we add it here or provide a single account on init?

chat/chat-invite.md
```diff
@@ -0,0 +1,38 @@
...
+
+A sends invite unecrypted to the invite topic.
+
+Invite topic is derived as the hash of the publicKey X.
+
+B decrypts opening message with the privateKey Y and deriving DH symKey I.
+
+B accepts the invite and generates a keyPair Z for chat thread.
+
+B sends response with publicKey Z on invite topic.
```
llbartekll: the question is whether the response should be encrypted(with symKeyI) or not? 
on one hand it is "cheap" to do so 
on the other - the request is not encrypted, so for consistency, the response could not be as well

chat/client-api.md
```diff
@@ -0,0 +1,62 @@
...
+  // ---------- Methods ----------------------------------------------- //
+
+  // initializes the client with persisted storage and a network connection
+  public abstract init(): Promise<void>;
+
+  // register a blockchain account with a public key / returns the public key
+  public abstract register(params: {
+    account: string;
+  }): Promise<string>;
+
```
llbartekll: Currently we have only one register() method that registers at a keyserver
but we are missing a method for "registering" for private discoverability, 
method that will create key pair, subscribes for private topic etc. and return pubKey

just a question if it is an MVP scope?

chat/client-api.md
```diff
@@ -0,0 +1,62 @@
...
+  // queries the default keyserver with a blockchain account / returns the public key
+  public abstract resolve(params: {
+    account: string;
+  }): Promise<string>;
+
+  // sends a chat invite with opening message / returns an invite id
+  public abstract invite(params: {
+    peerPublicKey: string;
+    invite: Invite;
+  }): Promise<number>;
```
llbartekll: we are inviting with a peerPublicKey - that method will work for all types of discoverability but we probably should add a label or an Account as an argument so inviter can associate something human readable with it's peer.

chat/chat-invite.md
llbartekll:
```suggestion
B sends response with publicKey Z on invite topic. Response is encrypted with symKey I.
```

chat/data-structures.md
```diff
@@ -0,0 +1,26 @@
...
+Invite is an encrypted payload used in wc_chatInvite which will be tracked by the json-rpc id to index on the client-side with the following params.
+
+```jsonc
+{
+  "message": string,
+  "account": string,
+  "signature": string, // optional
+}
+```
+
```
llbartekll, line 16: I am suggesting to use this data structure as it is inconvenient to keep pubkey and invite separately 
we can think of `pubkey` as of envelope address and of `invite` as an envelope content.
that structure could be a param in api's `invite()` method and emitted in `chat_invite` event.
We could also think of it as of opened envelope and of:
```jsonc
{
   pubKey: string,
   invite: string
}
```
as closed envelope 

```suggestion

### Invite Envelope
```jsonc
{
   pubKey: string,
   invite: {
     "message": string,
     "account": string,
      "signature": string, // optional
   }
}

```

chat/client-api.md
pedrouid: The Invite struct already includes the account as a parameter

chat/chat-invite.md
pedrouid: Yeah I feel like it's a weird design decision to not encrypt a request and then encrypt the response. That being said it's also more privacy-preserving because you wouldn't be able acceptance from rejection

chat/client-api.md
pedrouid: That's a good point. Perhaps, we could add a method for "generatePublicKey()"?

chat/chat-invite.md
pedrouid: So conflicted with this one 😅

chat/client-api.md
llbartekll: I think the method should return chat URI - similarly to `Sign.connect()`

 what about something like `registerPublicDiscoverability()` and `registerPrivateDiscoverability`

I think we could also introduce discoverability object, so user can opt in/out and track where is known as who.

chat/client-api.md
llbartekll: true, account of an inviter but I mean invitee account on inviter side so inviter can later associate that pubKey with someone.

chat/client-api.md
```diff
@@ -0,0 +1,62 @@
...
+
+  // accepts a chat invite by id / returns thread topic
+  public abstract accept(params: {
+    inviteId: string;
+  }): Promise<string>;
+
```
llbartekll, line 31:
```suggestion

  public abstract getMessages(params: {
    threadId: string;
  }): Promise<[Messages]>;
  
 public abstract getThreads(): Promise<[Threads]>;
  
```

make sense?

bkrem: Agreed, we currently don't specify here how the relevant entities (messages, threads, invites) are resolved. I guess it's implied this would happen in `init` right now and then the listeners - e.g. `on("chat_message",...)` - would take it from there?

So additionally `public abstract getInvites()` to cover all handled data structures?

chat/rpc-methods.md
pedrouid: I've removed it

Elyniss (approved):

chat/rpc-methods.md
```diff
@@ -50,7 +51,6 @@ Used to send a message to its peer through topic T.
 ```jsonc
 // wc_chatMessage params
 {
-  "publicKey": string,
```
bkrem: So we don't need the publicKey here because the message can be routed based on the thread topic anyways right?

bkrem (approved):

PR #75, opened 2022-06-27
add relay-network.md

pedrouid:
Add document defining schema serialization for inter-relay messages (aka iridium message)


relay/relay-network.md
```diff
@@ -0,0 +1,32 @@
...
+
+Relay network is defined by the federation of relay servers exchanging messages to route from a publishing client to a subscribing client on the same topic.
+
+## Goals
+
+Relay servers will route messages through other servers in the network that also have matching subscribers on a topic whilst maintaing the same expected behavior described by a publisher when requested to its own server
+
+## Iridium Message
+
+An Iridium message is defined as a serialized message of a successfully cached message from a publishing request from a connected client with one of the servers.
```
pedrouid: Should a topic also be encoded into this schema or we expect it to be known between the relay servers?

relay/relay-network.md
```diff
@@ -0,0 +1,32 @@
...
+
+- **prefix** - a string prefix ("irn") for exchanged iridium messages.
+- **version** - an integer number of the serialized schema for messages.
+- **tag** - an integer label that identifies a message without knowing its contents.
```
jakubuid, line 22: Change from api to tag should be applied on the client side as well in the Policy object

relay/relay-network.md
```diff
@@ -0,0 +1,67 @@
...
+
+- **topic** - a string defining the target topic for message subscription.
+- **hash** - a string defining the message id as the sha256 hash of itself.
+
+Header:
+
+- **prefix** - a string prefix ("irn") for exchanged iridium messages.
+- **version** - an integer number of the serialized schema for messages.
+- **tag** - an integer label that identifies a message without knowing its contents.
+- **prompt** - a flag that identifies whether the server should prompt a user action.
```
llbartekll, line 23: Just wonder if the `prompt` flag is still necessary when we have tags. Could relay server decide wether to prompt or not  base on tag value now? Or we want to leave that control to the client and potential 3rd party network consumers?

arein:
@pedrouid what part of the Relay server would adhere to that spec? Currently we have a HTTP interface to accept Relay messages - is that what this should cover?

relay/relay-network.md
```diff
@@ -0,0 +1,67 @@
...
+Relay servers will route messages through other servers in the network that also have matching subscribers on a topic whilst maintaing the same expected behavior described by a publisher when requested to its own server
+
+## Network Message
+
+An Network message is defined as a serialized message of a successfully cached message from a publishing request from a connected client with one of the servers.
+
+Context:
+
+- **topic** - a string defining the target topic for message subscription.
+- **hash** - a string defining the message id as the sha256 hash of itself.
```
sbc64, line 16: I think we should define this hash with an additional time value. I bumped into an error with redis recently with the jsonrpc tests in the ts-relay where two messages with the value of "SECOND_MESSAGE" would store in the same hash of redis in the message table.

My solution was to add an additional field to the message identifier: hash(topic, "SECOND_MESSAGE"). This ensure that each test run would make a new unique message since the topic was random.

I know that all message payloads are encrypted to therefore they are random enough and therefore we can ensure the uniqueness of each message. The problem occurs is that if the message payloads are not encrypted and are instead plain text one can send two messages with the same value and same topic and they would end up having the same `hash`.

I would trust more the output of the `hash` field if it contained some sort of salt. My suggestion is that we use `sha156(payload, jsonrp_id)` since the jsonrpcid is currently a semi random timer field. This is not a  perfect salt but at least now we can store **IF NEEDED** two identical payloads but sent at different times. 

Any thoughts on the suggestion? It is merely because I want to avoid bumping into a bug where we actually want to send two payloads identical payloads in a single topic.

sbc64: There is a risk that the network could be spammed with identical payloads but different `jsonrpcid`, but the network could still be spammed with spam/random message payloads.

relay/relay-network.md
sbc64: I think that the prompt is useful for triggering webhooks. Knowing when do to the webhook is something that the wallet knows and the relay can't know this. The effort on the echo server should clarify whether this is more or less useful.

pedrouid:
Closing this since it's been stagnant for too long

PR #86, opened 2022-07-13
Iridium params for Sign and Chat

### LINKED ISSUE #84: Granular Tags
Right now we have implemented iridium_publish tags as an integer that identifies each API: Unknown(0), Sign (1), Chat (2), Auth (3), Push (4)

However we want to move into more Granular tags per each JSON-RPC method to improve our network metrics to ensure users can successfully complete API-specific flows

This would look as follows for Sign API and Chat API:

------ SIGN -----

* wc_pairingDelete -> req: 1000 / res: 1001
* wc_pairingPing -> req: 1002 / res: 1003
* wc_sessionPropose -> req: 1100 / res: 1101
* wc_sessionSettle -> req: 1102 / res: 1103
* wc_sessionUpdate -> req: 1104 / res: 1105
* wc_sessionExtend -> req: 1106 / res: 1107
* wc_sessionRequest -> req: 1108 / res: 1109
* wc_sessionEvent -> req: 1110 / res: 1111
* wc_sessionDelete -> req: 1112 / res: 1113
* wc_sessionPing -> req: 1114 / res: 1115

------ CHAT -----

* wc_chatInvite -> req: 2000 / res: 2001
* wc_chatMessage -> req: 2002 / res: 2003
* wc_chatLeave -> req: 2004 / res: 2005
* wc_chatPing -> req: 2006 / res: 2007

These should be part of the spec of each API so that Relay and SDK are in sync

Todo:

- [x] https://github.com/WalletConnect/walletconnect-specs/pull/86
- [x] https://github.com/WalletConnect/WalletConnectKotlinV2/issues/249
- [x] https://github.com/WalletConnect/walletconnect-monorepo/issues/1242
- [x] https://github.com/WalletConnect/WalletConnectSwiftV2/issues/330
- [x] https://github.com/WalletConnect/ts-relay/issues/213
- [ ] https://github.com/WalletConnect/rs-relay/issues/122

pedrouid:
**NOTE:** This one is quite the PR that needs to be reviewed carefully.

Summary:
Introduction of granular tags and also specifying prompt and ttl for each RPC method.

Changes:
- Remove old tags for each API (0 = unknown,1 = Sign, 2 = Chat, 3 = Auth, 4 = Push) 
- Add wider range of tags per each RPC method (1xxx = Sign, 2xxx = Chat)
- Remove special cases for Publish TTL (-1 = indefinite, 0 = no caching)
- Add TTL values for each RPC method (30 = 30 secs, 300 = 5mins, 86400 = 24 hours)
- Add prompt flag for each RPC method (only for requests, never responses)

Some patterns that I believe we should follow for consistency when defining RPC methods:
- Request and Response SHOULD always match the same TTL
- Request CAN use Prompt flag but Response SHOULD be false
- Ping methods MUST use TTL of 30 secs
- Sign methods requiring user-action MUST use TTL of 5mins 
- Chat methods MUST use TTL of 24 hours
- Response tag SHOULD always be an increment of 1 matching Request tag
- Response success result SHOULD be empty params (instead of boolean)

Regarding TTL:
For the time being TTL has been set to be only up to 24 hours, given discussions with the Cloud team about expanding the infrastructure to support higher TTLs. In the future we might increase TTL of some methods especially for Chat API.

Issue:
Fix #84

### DIFF
--- a/chat/rpc-methods.md
+++ b/chat/rpc-methods.md
@@ -25,6 +25,8 @@ Used to invite a peer through topic I. Requires a success response with associat
 - Success response is equivalent to invite acceptance.
 - Error response is equivalent to invite rejection.
 
+**Request**
+
 ```jsonc
 // wc_chatInvite params
 {
@@ -33,13 +35,28 @@ Used to invite a peer through topic I. Requires a success response with associat
   "publicKey": string,
   "signature": string, // optional
 }
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | true     |
+| Tag     | 2000     |
+
 ```
+
+**Response**
 
 ```jsonc
 // Success result
 {
   "publicKey": string, // invitee public key
 }
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | false    |
+| Tag     | 2001     |
 ```
 
 ### wc_chatMessage
@@ -49,6 +66,8 @@ Used to send a message to its peer through topic T.
 - Success response is equivalent to message delivery receipt.
 - Error response is equivalent to message delivery failure.
 
+**Request**
+
 ```jsonc
 // wc_chatMessage params
 {
@@ -57,40 +76,93 @@ Used to send a message to its peer through topic T.
   "timestamp": Int64,
   "media": Media // optional
 }
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | true     |
+| Tag     | 2002     |
 ```
+
+**Response**
 
 ```jsonc
 // Success result
 {
   // empty
 }
-```
 
-### wc_chatPing
-
-Used to evaluate if peer is currently online. Timeout at 30 seconds
-
-```jsonc
-// wc_chatPing params
-{
-  // empty
-}
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | false    |
+| Tag     | 2003     |
 ```
 
 ### wc_chatLeave
 
 Used to signal to a peer that a chat thread is being left.
 
+**Request**
+
 ```jsonc
-// wc_chatLeave params
+// wc_chatPing params
 {
   // empty
 }
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | false    |
+| Tag     | 2004     |
 ```
+
+**Response**
 
 ```jsonc
 // Success result
 {
   // empty
 }
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | false    |
+| Tag     | 2005     |
+```
+
+### wc_chatPing
+
+Used to evaluate if peer is currently online. Timeout at 30 seconds
+
+**Request**
+
+```jsonc
+// wc_chatPing params
+{
+  // empty
+}
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 30       |
+| Prompt  | false    |
+| Tag     | 2006     |
+```
+
+**Response**
+
+```jsonc
+// Success result
+{
+  // empty
+}
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 30       |
+| Prompt  | false    |
+| Tag     | 2007     |
 ```
--- a/relay/relay-server.md
+++ b/relay/relay-server.md
@@ -10,9 +10,9 @@ The following definitions are shared concepts across all JSON-RPC methods for th
 
 - **topic** - a target topic for the message to be subscribed by the receiver.
 - **message** - a plaintext message to be relayed to any subscribers on the topic.
-- **ttl** - a storage duration for the message to be cached server-side in **seconds** (aka time-to-live). **(0 - no caching, -1 - forever)**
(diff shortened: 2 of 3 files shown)


chat/rpc-methods.md
```diff
@@ -57,40 +76,93 @@ Used to send a message to its peer through topic T.
...
-Used to evaluate if peer is currently online. Timeout at 30 seconds
+Used to signal to a peer that a chat thread is being left.
+
+**Request**
 
 ```jsonc
 // wc_chatPing params
 {
   // empty
 }
```
jakubuid:
```suggestion
// wc_chatLeave params
{
  // empty
}
```

chat/rpc-methods.md
```diff
@@ -57,40 +76,93 @@ Used to send a message to its peer through topic T.
...
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Prompt  | false    |
```
jakubuid: Are we sure setting prompt to false for wc_chatLeave?
For example when the other peer's app is in the background  and the second peer decides to leave, the first one doesn't know about it until putting app into foreground.

jakubuid:
> * Response success result SHOULD be empty params (instead of boolean)

@pedrouid I'm just curious could you elaborate why this change? 

By empty you mean null or empty string?
1. `{"jsonrpc": "2.0", "result": "", "id": 1}`
2. `{"jsonrpc": "2.0", "result": null, "id": 1}`

chat/rpc-methods.md
pedrouid: Fair point! 👍 I will change it

pedrouid:
> > * Response success result SHOULD be empty params (instead of boolean)
>
> By empty you mean null or empty string?

I mean... I'm also still slightly debating on this because there is a mix of empty params and boolean results between Sign and Chat

```
// wc_chatPing params
{
   // empty
}

// Success result
true
```

So it would be between the bottom two options:

1. `{"jsonrpc": "2.0", "result": {}, "id": 1}`
2. `{"jsonrpc": "2.0", "result": true, "id": 1}`

I guess it would be less disruptive to just the option 2

pedrouid:
OK @jakobuid, I've just changed all success result without params to have a boolean true... essentially they are acknowledgements

PS - also updated PR description to remove confusion

jakobuid (approved):

llbartekll (approved):

relay/relay-server.md
```diff
@@ -10,9 +10,9 @@ The following definitions are shared concepts across all JSON-RPC methods for th
 
 - **topic** - a target topic for the message to be subscribed by the receiver.
 - **message** - a plaintext message to be relayed to any subscribers on the topic.
-- **ttl** - a storage duration for the message to be cached server-side in **seconds** (aka time-to-live). **(0 - no caching, -1 - forever)**
-- **tag** - a label that identifies which api sent a message. It allows collecting metrics. **(0 - unknown, 1 - sign, 2 - chat, 3 - auth, 4 - push)**
-- **prompt** - a flag that identifies whether the server should trigger a notification webhook to a client through a push server **(true/false)**
+- **ttl** - a storage duration for the message to be cached server-side in **seconds** (aka time-to-live).
```
arein, line 13: As discussed i think deriving the TTL form the tag is the better option rather than letting the client define this on a per message basis. As a user of the walletconnect protocol, this extra cognitive load confuses me. I'd rather have it implicit and only worry about the tags.

This also opens up more DDOS avenues.

It's ultimately not in line with our principle of being developer friendly.

arein (approved):

relay/relay-server.md
pedrouid: TLDR' we can prevent DDOS avenues by setting TTL to a maximum of 24 hours as described above

-----------------------------------------------------------------------------

I still strongly disagree with this as I mentioned in Koh Samui because it does not affect developer-experience, the Iridium RPC sits at a very low-level of the stack

Sign/Chat SDKs -> Relay Client -> Iridium RPC -> Relay Server

Developers will only interface with Sign/Chat SDKs which will control the TTL's for each RPC method and will interface with the Relay Client, it's only at the Relay Client internally that Iridium RPC are formatted to comply with Relay Server requirements

Additionally we have made TTL required because it affects the operation of publishing messages in the network and as described above there isn't a single TTL that is expected to be higher than 24 hours (86400 seconds) which was what we discussed would be our current guarantees provided by Redis-based architecture.

Another reason to not rely on tags for setting TTL at the server level is because this strongly couples Sign and Chat behavior with the Relay Server... this means everytime any of the consumer APIs change behavior then it needs coordination with the Relay Server which makes the whole system way less modular.

Finally tags are meant to always be optional for backwards-compatibility and privacy-perserving. While WalletConnect Cloud will require tagging for publishing messages to not affect our business metrics, any self-host instance is able to disable tagging on their end while still being fully operational with the Iridium network.

chadyj (approved):

sign/rpc-methods.md
```diff
@@ -18,7 +18,74 @@ error: {
...
+```jsonc
+// wc_pairingPing params
+{
+  // empty
+}
+
+| Iridium |          |
+| ------- | -------- |
+| TTL     | 30       |
+| Prompt  | true     |
```
bkrem: We have prompt `true` here but false on `wc_sessionPing` and for the chat ping as far as I can tell. Is this intentional here?

bkrem (approved):

sbc64 (approved):

sign/rpc-methods.md
pedrouid: Great catch!

PR #88, opened 2022-07-18
Auth API

pedrouid:
First draft to Auth API spec

Related to #77


auth/request-uri.md
```diff
@@ -0,0 +1,33 @@
...
+    parameters  = parameter *( "&" parameter )
+    parameter   = key "=" value
+    key         = STRING
+    value       = STRING
+
+## Parameters
+
+### Required
+
+- `symKey` (STRING) = account associated with public key
```
llbartekll: > account associated with public key

what does it mean? typo?

auth/request-uri.md
```diff
@@ -0,0 +1,33 @@
...
+- `relay-protocol` (STRING) = protocol name used for relay
+
+### Optional
+
+- `relay-data` (STRING) = hex data payload used for relay
+
+## Example
+
+    topic = “7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9”
```
llbartekll, line 26: > Pairing topic is the hash of symKey S.

we could probably remove topic as it is hash of symKey?
on the other hand I can see it is not true for SignAPI so probably we should stay consistent?

auth/rpc-methods.md
```diff
@@ -0,0 +1,103 @@
...
+      "name": string,
+      "description": string,
+      "url": string,
+      "icons": [string]
+    },
+  },
+  "header": {
+    "t": string,
+  },
+  "payload": {
```
llbartekll: maybe header and payload should not be flat with `requester`?
wrap into `CACAO`?

auth/client-api.md
```diff
@@ -0,0 +1,22 @@
...
+    chainId: string;
+    nonce: string;
+    aud?: string;
+    nbf?: string;
+    exp?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+  }): Promise<boolean>;
+}
```
llbartekll, line 32:
```suggestion
// for wallet to listen on auth request
  public abstract on("on_auth_request", (topic: string, cacao: UnsignedCACAO) => {}): void;
  }
```

auth/client-api.md
```diff
@@ -0,0 +1,22 @@
...
+  public abstract authenticate(params: {
+    chainId: string;
+    nonce: string;
+    aud?: string;
+    nbf?: string;
+    exp?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+  }): Promise<boolean>;
```
llbartekll: could params be `requestId` aka `requester.publicKey` or topic  and `SignedCACAO` ?
I assume this method is for wallet.

auth/client-api.md
```diff
@@ -0,0 +1,22 @@
+# Client API
+
+```typescript
+abstract class Client {
+  // ---------- Methods ----------------------------------------------- //
+
+  // initializes the client with persisted storage and a network connection
```
llbartekll, line 7: should we add following methods:
`requestAuthentication()` - for app that creates a pairing and sends `wc_authRequest`, 
`pair(`) ?
`ping()` ?

auth/authentication.md
```diff
@@ -0,0 +1,61 @@
...
+
+Response topic is the hash of publicKey X.
+
+A will construct an authentication request.
+
+A encrypts request with symKey S.
+
+A publishes encrypted request to topic.
+
+A starts polling messages on response topic.
```
llbartekll, line 43: polling? - so we are planning to not have a websocket on A?

auth/authentication.md
```diff
@@ -0,0 +1,61 @@
...
+User returns to a previously visted website before authentication expiry
+
+1. Website checks if user is authenticated from local storage
+2. Website is now authenticated with signed message
+
+User returns to a previously visted website after authentication expiry
+
+1. Website checks if user is authenticated from local storage
+2. Website sends new request to previously paired wallet
+3. User views notification and redirects to wallet
```
jakubuid: If I understand correctly would it require a push support on mobile devices?

auth/README.md
```diff
@@ -3,3 +3,23 @@
...
+
+Auth API is a one-click passwordless authentication provider for any website or application.
+
+## Goals
+
+A user can login or sign-in into any website or application with their blockchain account.
+
+An app does not need to maintain a persisted connection with the wallet once authenticated.
+
+A wallet can evaluate all auth parameters requested to format the message independently.
```
bkrem, line 23: By `independently` do we here mean simply that the Auth API does not handle the message formatting, or that that there won't be a (suggested) standard for authentication message formatting, i.e. wallets can implement this as they wish?

auth/authentication.md
jakubuid: By polling you mean a GET HTTP where topic is added as query?
Or POST HTTP with JsonRpcRequest as body where topic is inside params?

auth/authentication.md
```diff
@@ -0,0 +1,61 @@
...
+
+A encrypts request with symKey S.
+
+A publishes encrypted request to topic.
+
+A starts polling messages on response topic.
+
+B subscribes to pairing topic from provided URI
+
+B receives request and decrypts it with symKey S.
```
jakubuid: It's prob obvious but I'd add that symKey S is retrieved from URI.

auth/authentication.md
```diff
@@ -0,0 +1,61 @@
...
+
+B constructs message to be signed from request and signs it.
+
+B generates keyPair Y and generates shared symKey R.
+
+B encrypts response with symKey R as type 1 envelope.
+
```
jakubuid, line 54: Looks like line how response topic is created is missing.

`Response topic is the hash of pubKey Y.`

Is that correct?

auth/client-api.md
jakubuid: I agree

jakobuid (changes requested):

auth/rpc-methods.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+### wc_authRequest
+
+Used to request authentication signature to a peer through topic P. Response is expected on topic R with shared symKey.
+
+Parameters for request and response are compliant with [CAIP-74](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-74.md) (aka CACAO)
+
+- Success response is equivalent to authentication acceptance.
```
jakubuid, line 27: Imho it should be more detailed: `Success response is equivalent to authentication acceptance only if signature is valid.`

auth/request-uri.md
pedrouid: typo... will fix

auth/request-uri.md
pedrouid: technically we could but we should aim for consistency

auth/authentication.md
pedrouid: Correct... we need to coordinate this with Iridium team to include a `irn_getMessages` endpoint

auth/README.md
pedrouid: Correct, given the payload type the wallet should be able to format the message before signing instead of expecting the dapp to format it

auth/authentication.md
pedrouid: Not necessarily a requirement because you have a websocket on the wallet side but it does improve the UX

auth/rpc-methods.md
pedrouid: I think it should be validated on the dapp side as well

auth/client-api.md
```diff
@@ -0,0 +1,24 @@
...
+  public abstract init(): Promise<void>;
+
+  // for responder to pair a pairing created by a proposer
+  public abstract pair(params: {
+    uri: string;
+  }): Promise<Sequence>;
+
+  // request wallet authentication
+  public abstract request(params: RequestParams): Promise<boolean>;
```
llbartekll:
```suggestion
  // request wallet authentication
  // returns URI if pairing topic not provided
  public abstract request(params: RequestParams, pairingTopic?: string): Promise<string?>;
```

auth/client-api.md
```diff
@@ -0,0 +1,24 @@
...
+
+  // respond wallet authentication
+  public abstract respond(params: RespondParas): Promise<boolean>;
```
llbartekll:
```suggestion
  public abstract respond(params: RespondParams): Promise<boolean>;
```

auth/authentication.md
flypaper0: Websockets / long polling require app to be in foreground. Push notifications is the only way to make it properly. Otherwise we can ask user to scan QR again

pedrouid: The pattern isn't that different from Sign API.

User is either prompted or redirected to the Wallet to approve the authentication signature

The Push Notification would be a nice-to-have

However I can see how it's adding confusion to the spec... I will rephrase it

auth/authentication.md
pedrouid: It's mentioned on Line 35

Might need to restructure how we write these protocol flows

auth/client-api.md
pedrouid: I mean... this is a weird one to spec out because the Sign API also doesn't have this specced out

However the Javascript pattern is two have either a URI as an output or not and additionally a Promise to be resolved

I added a pairing_created to the spec like Sign API for consistency

auth/data-structures.md
```diff
@@ -0,0 +1,92 @@
...
+interface RequestParams {
+  chainId: string;
+  nonce: string;
+  aud?: string;
+  nbf?: string;
+  exp?: string;
+  statement?: string;
+  requestId?: string;
+  resources?: string[];
+}
```
llbartekll, line 15:
```suggestion
interface RequestParams {
  domain: string;
  chainId: string;
  nonce: string;
  aud: string;
  nbf?: string;
  exp?: string;
  statement?: string;
  requestId?: string;
  resources?: string[];
}
```

llbartekll: from specs:
uri in siwe is not optional so aud should not be as well
`URI: {.p.aud}`
domain is required I think

auth/data-structures.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+## Respond Params
+
+```typescript
+interface RespondParams {
+  topic: string;
+  signature: CacaoSignature;
```
llbartekll, line 23: I think signature may not be enough as requester do not know the whole payload and type.
could we return Cacao instead?

PR #92, opened 2022-07-25
Update codes.md

Elyniss:
docs: Add session request reject code


Elyniss:
Decided to only add `User Rejected Request` from [eip-1193](https://eips.ethereum.org/EIPS/eip-1193) because others are already supported with other error codes

jakobuid (approved):

PR #97, opened 2022-08-02
Update data-structures.md

llbartekll:
`auth_request` event provides formatted eip-4361 message and request id, so I think it is more convenient and makes more sense that `respondParams` contains associated request id in it's body.


llbartekll:
actually response topic is also unique so it does not matter

llbartekll:
Reopening xD, It's more convenient to manage by id.

llbartekll:
Reopening xD, It's more convenient to manage by id.

PR #98, opened 2022-08-03
Update data-structures.md

llbartekll:
in caip-74 specs chainID is derived from issuer's caip-10:
`Chain ID: {.p.iss[chainId]}`
so maybe it make sense that chain id is set on issuer's side when responding?


pedrouid:
This should be a dapp requirement

PR #103, opened 2022-08-11
Update client-api.md

llbartekll:
As partially discussed on Wednesday, wallet should be able to respond with error.
auth_response event should also be able to emit with error - error could happen either or error response or on EIP-4361 validation on app side.

### DIFF
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -25,7 +25,7 @@ abstract class Client {
   // ---------- Events ----------------------------------------------- //
 
   // subscribe to auth response
-  public abstract on("auth_response", (id: number, result: Result<Address, Error>) => {}): void;
+  public abstract on("auth_response", (id: number, result: Address | Error) => {}): void;
 
   // for wallet to listen on auth request
   public abstract on("auth_request", (id: number, message: string) => {}): void;
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -20,12 +20,12 @@ abstract class Client {
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
 
   // query cached request matching id
-  public abstract getRequest(params: { id: number }): Promise<Cacao>;
+  public abstract getRequest(params: { id: number }): Promise<Cacao | Error>;
 
   // ---------- Events ----------------------------------------------- //
 
   // subscribe to auth response
-  public abstract on("auth_response", (id: number, result: Address | Error) => {}): void;
+  public abstract on("auth_response", (id: number, result: Cacao | Error) => {}): void;
 
   // for wallet to listen on auth request
   public abstract on("auth_request", (id: number, message: string) => {}): void;
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -14,7 +14,7 @@ abstract class Client {
   public abstract request(params: RequestParams): Promise<{ uri, id }>;
 
   // respond wallet authentication
-  public abstract respond(params: RespondParams): Promise<boolean>;
+  public abstract respond(result: Resutl<RespondParams, Error>): Promise<boolean>;
 
   // query all pending requests
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
@@ -25,7 +25,7 @@ abstract class Client {
   // ---------- Events ----------------------------------------------- //
 
   // subscribe to auth response
-  public abstract on("auth_response", (id: number, cacao: Cacao) => {}): void;
+  public abstract on("auth_response", (id: number, result: Result<Address, Error>) => {}): void;
 
   // for wallet to listen on auth request
   public abstract on("auth_request", (id: number, message: string) => {}): void;
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -14,7 +14,7 @@ abstract class Client {
   public abstract request(params: RequestParams): Promise<{ uri, id }>;
 
   // respond wallet authentication
-  public abstract respond(result: Result<RespondParams, Error>): Promise<boolean>;
+  public abstract respond(result: RespondParams | Error): Promise<boolean>;
 
   // query all pending requests
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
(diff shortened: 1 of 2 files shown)


auth/client-api.md
```diff
@@ -14,7 +14,7 @@ abstract class Client {
   public abstract request(params: RequestParams): Promise<{ uri, id }>;
 
   // respond wallet authentication
-  public abstract respond(params: RespondParams): Promise<boolean>;
+  public abstract respond(result: Resutl<RespondParams, Error>): Promise<boolean>;
```
Elyniss:
```suggestion
  public abstract respond(result: Result<RespondParams, Error>): Promise<boolean>;
```

llbartekll: didn't ask for review yet 🤣 
I think it is not javascript btw

llbartekll: but commited xd

auth/client-api.md
```diff
@@ -25,7 +25,7 @@ abstract class Client {
   // ---------- Events ----------------------------------------------- //
 
   // subscribe to auth response
-  public abstract on("auth_response", (id: number, cacao: Cacao) => {}): void;
+  public abstract on("auth_response", (id: number, result: Result<Address, Error>) => {}): void;
```
llbartekll:
```suggestion
  public abstract on("auth_response", (id: number, result: Address | Error) => {}): void;
```

auth/client-api.md
```diff
@@ -14,7 +14,7 @@ abstract class Client {
   public abstract request(params: RequestParams): Promise<{ uri, id }>;
 
   // respond wallet authentication
-  public abstract respond(params: RespondParams): Promise<boolean>;
+  public abstract respond(result: Result<RespondParams, Error>): Promise<boolean>;
```
llbartekll:
```suggestion
  public abstract respond(result: RespondParams | Error): Promise<boolean>;
```

auth/client-api.md
```diff
@@ -25,7 +25,7 @@ abstract class Client {
   // ---------- Events ----------------------------------------------- //
 
   // subscribe to auth response
-  public abstract on("auth_response", (id: number, cacao: Cacao) => {}): void;
+  public abstract on("auth_response", (id: number, result: Address | Error) => {}): void;
```
bkrem: Why is the success type `Address` here?

Elyniss: I think it's better to stay with Cacao

auth/client-api.md
```diff
@@ -14,7 +14,7 @@ abstract class Client {
   public abstract request(params: RequestParams): Promise<{ uri, id }>;
 
   // respond wallet authentication
-  public abstract respond(params: RespondParams): Promise<boolean>;
+  public abstract respond(result: RespondParams | Error): Promise<boolean>;
```
jakubuid, line 17: Is `Error` structure defined somewhere? I know that's JsonRpcError but it'd be nice to mention it

auth/client-api.md
llbartekll: @bkrem  of Address I thought as of typealias for String.

@Elyniss why you think it is better to stay with cacao?
I think the only info dapp needs is an address if authentication was successful. Additionally cacao issuer is didpkh so it quite inconvenient to derive an address. Plus if app needs some data from request/response it is suppose to be able to query request by id from a client

auth/client-api.md
llbartekll: I do not think it is a jsonrpc error, it is more like an aggregation of jsonrpc error and internal signature validation error

jakubuid: So it's even more required to have this structure defined somewhere

llbartekll: Not necessarily a structure but I think we could agree on error codes

llbartekll: I know what you mean:
```
Error {
code: number
message: string
}
```

llbartekll: yeah we can add it if you guys accept the PR

auth/client-api.md
Elyniss: I agree that for use cases 99% only `address` would suffice, but what if dapp would like some data about `nbf` and `exp`?

llbartekll: `nbf` and `exp` is set by dapp so dapp is suppose to know it and can querry it by calling:

```
public abstract getRequest(params: { id: number }): Promise<Cacao>;
```

actually I would change the return value as well for something more precise, maybe is cacao valid?

Elyniss: Yeah, you're right about `nbf` and `exp`. I agree then only returning address would suffice.

However, I would not change this `public abstract getRequest(params: { id: number }): Promise<Cacao>;`. If we keep it we don't force dapps to store whole cacao as only id would suffice. But I don't have strong opinion on this. I only think that leaving it is a better dev experience.

llbartekll: > Yeah, you're right about nbf and exp. I agree then only returning address would suffice.

Now I do not agree with you 🤣  and I think you had a good point....
I suppose dapp is going to construct jwt from cacao properties so it may be useful to have cacao in the event 🤔

llbartekll: > However, I would not change this public abstract getRequest(params: { id: number }): Promise<Cacao>;. If we keep it we don't force dapps to store whole cacao as only id would suffice. But I don't have strong opinion on this. I only think that leaving it is a better dev experience.

I think you misunderstood, I meant that cacao may not be enough 🤔 
because what if response was an error or signature validation failed? should we return invalid cacao with this function without telling is invalid? something else?

llbartekll: anyway:
```
  public abstract on("auth_response", (id: number, result: (cacao, string) | Error) => {}): void;
```
where string is an address?
too verbose?

Elyniss: I'm still more favouring the whole Cacao.

llbartekll: to which `auth_response` ?

Elyniss (approved):
What about Error. Should we add it do data-structures file?

llbartekll:
> What about Error. Should we add it do data-structures file?

@Elyniss  it's already there

llbartekll:
maybe we should also create codes.md file

bkrem (approved):

PR #104, opened 2022-08-16
Improve Auth API spec and data structures

pedrouid:
I like the improvements that you made @llbartekll to the Auth API responses but for consistency I've refactored some of the specs and data structures to improve consistency

Let me know what you think 🙏


sign/error-codes.md
```diff
@@ -1,4 +1,4 @@
-# Codes
+# Error Codes
```
Elyniss, line 1: We renamed Error Codes to Codes because we've added `case .userDisconnected: return 6000`, which is technically not an error.

pedrouid: Remember we actually used to call these Reason Codes before beta.100... I can't remember why we stopped calling them like that

jakobuid (commented):
LGTM

llbartekll (approved):
looks good!
that was me with javascript so I expected correction :D

sbc64 (approved):

auth/data-structures.md
```diff
@@ -37,10 +37,7 @@ interface RequestParams {
 ## Respond Params
 
 ```typescript
-interface RespondParams {
-  id: number;
-  signature: CacaoSignature;
-}
+type RespondParams = ResultResponse | ErrorResponse;
```
llbartekll, line 40: we also need an id for error response so maybe something like?:

```suggestion
type RespondParams = (id: number, result: ResultResponse | ErrorResponse);
```

Elyniss: What about using `id` param inside `ResultResponse`? Should we add `id` to `ErrorResponse` as well?

pedrouid: I added id to ErrorResponse instead since ResultResponse already has an id

sign/error-codes.md
pedrouid: Honestly let's go with this for now and then we can revisit

PR #105, opened 2022-08-22
[Auth] Error codes

flypaper0:


auth/codes.md
```diff
@@ -0,0 +1,16 @@
+# Codes
+
+## REJECTED
+
+```sh
+case .userRejectedRequest return 4001
```
Elyniss: How about we use `1` before Auth specific codes?
```suggestion
case .userRejectedRequest return 14001
```

auth/codes.md
```diff
@@ -0,0 +1,16 @@
...
+```
+
+## VALIDATION
+
+```sh
+case .malformedResponseParams return 7001
+case .malformedRequestParams return 7002
+case .messageCompromised return 7003
+case .messageVerificationFailed return 7004
```
Elyniss:
```suggestion
case .malformedResponseParams return 12001
case .malformedRequestParams return 12002
case .messageCompromised return 12003
case .messageVerificationFailed return 12004
```

Elyniss (approved):

auth/codes.md
```diff
@@ -0,0 +1,16 @@
+# Codes
+
+## REJECTED
+
+```sh
+case .userRejectedRequest return 14001
```
llbartekll: I would reorder them so 14000s are after 12000s

auth/codes.md
```diff
@@ -0,0 +1,16 @@
...
+```
+
+## VALIDATION
+
+```sh
+case .malformedResponseParams return 12001
+case .malformedRequestParams return 12002
+case .messageCompromised return 12003
+case .messageVerificationFailed return 12004
```
llbartekll: does `signatureVerificationFailed` sounds better?

auth/codes.md
```diff
@@ -0,0 +1,16 @@
...
+## REJECTED
+
+```sh
+case .userRejectedRequest return 14001
+```
+
+## VALIDATION
+
+```sh
+case .malformedResponseParams return 12001
```
llbartekll: any reason why 12000s not 11000s ?

auth/codes.md
jakubuid: Why can't we inherit common error codes from sign? `userRejectedRequest` is already there, we could create a shared pool of errors and avoid creating the same errors with just different code.

Elyniss: I think we should distinguish between sessionRequest and authRequest

jakubuid: Why? For peers it doesn't matter. For Relay as well. Auth peer doesn't know about Sign peer.

Elyniss: I don't have a strong opinion on it. I just think it is not necessary to create a shared pool of errors just for one error.

jakubuid: Yea I agree that for only one error it doesn't make sense. Let's revisit this once more shared error occurs.

auth/codes.md
```diff
@@ -0,0 +1,16 @@
+# Codes
+
+## REJECTED
+
+```sh
+case .userRejectedRequest return 14001
+```
+
+## VALIDATION
```
llbartekll: still wondering if those errors really need codes and specs. They will be never send to the peer. 🤔

auth/codes.md
```diff
@@ -0,0 +1,16 @@
+# Codes
+
```
llbartekll, line 2:
```suggestion
## REASON
```sh
case .userDisconnected: return 6000
```

```

PR #107, opened 2022-08-22
Request Uri 

Elyniss:


IljaDaderko (approved):

sign/pairing-uri.md
```diff
@@ -2,7 +2,7 @@
 
 ## Schema ([EIP-1328](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1328.md))
 
-    uri         = "wc" ":" topic [ "@" version ][ "?" parameters ]
+    uri         = "wc" ":" "sign" "-" topic [ "@" version ][ "?" parameters ]
```
jakubuid, line 5: This change breaks backward compatibility in `sign`. We cannot be sure that all Dapps will use the new sdk in the same time.

We should either handle `sign` URI with prefix and without prefix on wallets side or just don't add such prefix and assume that when there is no prefix we consume `sign` sdk.

xzilja: what if this becomes uri param instead? i.e. `?sdk=sign` if there is no param we assume it is sign

Elyniss: That's why I've added Backwards Compatibility section

> This change breaks backward compatibility in `sign`. We cannot be sure that all Dapps will use the new sdk in the same time.
> 
> We should either handle `sign` URI with prefix and without prefix on wallets side or just don't add such prefix and assume that when there is no prefix we consume `sign` sdk.

jakubuid: Ahh sorry @Elyniss I didn't notice

sign/pairing-uri.md
```diff
@@ -29,5 +29,20 @@
     relay = { protocol: "irn", data: "" }
 
 ```
-uri = "wc:7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=irn&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303"
+uri = "wc:sign-7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=irn&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303"
+```
+
+## Backwards Compatibility
+1. URI MAY not contain `sign-` prefix to bo valid
```
jakubuid: `bo` ?

Elyniss:
```suggestion
1. URI MAY not contain `sign-` prefix to be valid
```

jakobuid (approved):

sign/pairing-uri.md
```diff
@@ -20,8 +20,11 @@
 ### Optional
 
 - `relay-data` (STRING) = hex data payload used for relay
+- `sdk` (STRING) = sdk that should handle this uri
 
-## Example
+## Examples
+
+1. Only required parameters
```
Talhaali00: maybe we call this the deprecated example since we want to move forward with the sdk parameter

Elyniss: Nice call, I like it

Elyniss: Also reordered them

pedrouid:
There is an issue with this approach because a URI like Chat API uses will have a different schema than the one used by Auth and Sign so if you only add the SDK on the query param (which is at the end) then you can't properly parse the URI

Elyniss:
Moved back to previous proposal without `sdk` parameter

auth/request-uri.md
```diff
@@ -29,5 +29,5 @@
     relay = { protocol: “iridium”, data: “” }
 
 ```
-uri = “wc:7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=iridium&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303”
+uri = “wc:auth-7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=iridium&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303”
```
devceline, line 32: Can we change `iridium` to `irn` here please?

devceline: Can we change `iridium` to `irn` here?

llbartekll: good catch, I think it should be changed

auth/request-uri.md
```diff
@@ -29,5 +29,5 @@
     relay = { protocol: “iridium”, data: “” }
```
jakubuid: And here as well `iridium` to `irn`

sign/pairing-uri.md
```diff
@@ -29,5 +29,20 @@
...
+uri = "wc:sign-7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=irn&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303"
+```
+
+## Backwards Compatibility
+1. URI MAY not contain `sign-` prefix to be valid
+
+### Schema
+
+```
+    uri         = "wc" ":" topic [ "@" version ][ "?" parameters ]
```
bkrem: Leading whitespace:

```suggestion
uri         = "wc" ":" topic [ "@" version ][ "?" parameters ]
```

sign/pairing-uri.md
```diff
@@ -29,5 +29,20 @@
...
+```
+
+### Example
+
+```
+    uri = "wc:7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=irn&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303"
```
bkrem: Leading whitespace (being rendered as dots in the markdown):

```suggestion
uri = "wc:7f6e504bfad60b485450578e05678ed3e8e8c4751d3c6160be17160d63ec90f9@2?relay-protocol=irn&symKey=587d5484ce2a2a6ee3ba1962fdd7e8588e06200c46823bd18fbd67def96ad303"
```

bkrem (approved):

pedrouid:
@Elyniss is this good to merge?

PR #108, opened 2022-08-25
Consolidate Core APIs into a single spec

pedrouid:
One section that covers three shared core modules into a single spec:
* Relay API
* Crypto API
* Storage API


Elyniss (approved):

core/README.md
```diff
@@ -0,0 +1,19 @@
+# Core API Overview
+
+## Description
+
+Core API consolidates several core modules that are shared between all other high-level APIs avilable to developers on WalletConnect 2.0
```
jakubuid:
```suggestion
Core API consolidates several core modules that are shared between all other high-level APIs available to developers on WalletConnect 2.0
```

core/crypto-api.md
```diff
@@ -0,0 +1,107 @@
...
+  ): Promise<string>;
+
+  public abstract decode(
+    topic: string,
+    encoded: string,
+    opts?: CryptoTypes.DecodeOptions
+  ): Promise<JsonRpcPayload>;
+
+  public abstract signJWT(aud: string): Promise<string>;
+}
```
jakubuid, line 102: I like the idea of having one place where we agree on the core components but for Android the code snippet above is not very useful. For example we have a separate class for JWT work, and we have more generic methods in regards of setting sym/pub keys. 
Should we just stick only to the general description of what given module is responsible for and drop those api examples?
Whether it makes sense for product apis (Sign, Auth, Chat), in here I'm not quite sure. 
This may also applies for swift team.

llbartekll: yeah I agree, on one hand it's good to generalise the concept so everyone knows what we are talking about on the other we already have different implementations so it's difficult to agree on same specs.

pedrouid: No worries, I will remove the JWT stuff for now

PR #109, opened 2022-08-30
Update Auth client-api.md

llbartekll:
As discussed on yesterday's sdk meeting.

### DIFF
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -19,7 +19,7 @@ abstract class Client {
   // query all pending requests
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
   
-  // querry pairings
+  // query pairings
   public abstract getPairings(): Promise<Array<Pairing>>;
   
   // for either to ping a peer in a pairing
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -26,7 +26,7 @@ abstract class Client {
   public abstract ping(params: { topic: string; }): Promise<void>;
   
   // for either peer to disconnect a pairing
-  public abstract disconnect(params: { topic: string, reason: Reason}): Promise<void>;
+  public abstract disconnect(params: { topic: string }): Promise<void>;
 
 
   // ---------- Events ----------------------------------------------- //
--- a/auth/client-api.md
+++ b/auth/client-api.md
@@ -25,7 +25,7 @@ abstract class Client {
   // for either to ping a peer in a pairing
   public abstract ping(params: { topic: string; }): Promise<void>;
   
-  // for either to disconnect a pairing
+  // for either peer to disconnect a pairing
   public abstract disconnect(params: { topic: string, reason: Reason}): Promise<void>;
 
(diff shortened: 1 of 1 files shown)


jakobuid (approved):

jakobuid (approved):

auth/client-api.md
```diff
@@ -18,9 +18,12 @@ abstract class Client {
...
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
-
-  // query cached response matching id
-  public abstract getResponse(params: { id: number }): Promise<Response>;
+  
+  // querry pairings
+  public abstract getPairings(): Promise<Array<Pairing>>;
+  
+  // for either to ping a peer in a pairing
+  public abstract ping(params: { topic: string; }): Promise<void>;
```
jakubuid, line 26: If we let users to get list of pairings should we also allow them to delete one? Imo It's be cool

llbartekll: make sense to me, I will add it!

auth/client-api.md
```diff
@@ -18,9 +18,16 @@ abstract class Client {
...
   // query all pending requests
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
+  
+  // querry pairings
+  public abstract getPairings(): Promise<Array<Pairing>>;
+  
+  // for either to ping a peer in a pairing
+  public abstract ping(params: { topic: string; }): Promise<void>;
+  
+  // for either to disconnect a pairing
```
bkrem:
```suggestion
  // for either peer to disconnect a pairing
```

auth/client-api.md
```diff
@@ -18,9 +18,16 @@ abstract class Client {
 
   // query all pending requests
   public abstract getPendingRequests(): Promise<Record<number, PendingRequest>>;
+  
+  // querry pairings
```
bkrem:
```suggestion
  // query pairings
```

auth/client-api.md
```diff
@@ -18,9 +18,16 @@ abstract class Client {
...
+  public abstract getPairings(): Promise<Array<Pairing>>;
+  
+  // for either to ping a peer in a pairing
+  public abstract ping(params: { topic: string; }): Promise<void>;
+  
+  // for either to disconnect a pairing
+  public abstract disconnect(params: { topic: string, reason: Reason}): Promise<void>;
```
bkrem: Should we add `Reason` to the data structures?

bkrem (approved):

auth/client-api.md
llbartekll: actually, the reason will be always .userDisconnected 
so what about:
```suggestion
  public abstract disconnect(params: { topic: string }): Promise<void>;
```

llbartekll: @bkrem ⬆️

bkrem: True since it’s an explicit top level action here it’s always user initiated 👍

PR #111, opened 2022-09-02
Update rpc-methods.md

llbartekll:
Those are duplicates to what we have in Sign but I would keep it this way until we extract pairing.

### DIFF
--- a/auth/rpc-methods.md
+++ b/auth/rpc-methods.md
@@ -67,7 +67,7 @@ Parameters for request and response are compliant with [CAIP-74](https://github.
 
 ### wc_pairingDelete
 
-Used to inform the peer to close and delete a pairing. All associated sessions of the given pairing must also be deleted.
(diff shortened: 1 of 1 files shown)


auth/rpc-methods.md
```diff
@@ -63,3 +63,69 @@ Parameters for request and response are compliant with [CAIP-74](https://github.
 | Prompt  | false    |
 | Tag     | 3001     |
 ```
+
+
+### wc_pairingDelete
+
+Used to inform the peer to close and delete a pairing. All associated sessions of the given pairing must also be deleted.
```
bkrem: `sessions` terminology doesn't work here, so maybe this instead?

```suggestion
Used to inform the peer to close and delete a pairing. The associated authentication state of the given pairing must also be deleted.
```

jakobuid (approved):

bkrem (approved):

PR #112, opened 2022-09-05
Revert Sign URI prefix changes

pedrouid:


pedrouid:
Moved to WalletConnect/walletconnect-docs#258

PR #117, opened 2023-08-08
Add Hedera RPC reference page

huck-hgraph:
### Summary
Adds RPC Reference page for Hedera with documentation of the following JSON-RPC methods:
- `hedera_signAndExecuteTransaction`
- `hedera_signAndReturnTransaction`
- `hedera_signMessage`


pedrouid:
Please make a PR to the new Specs repo: https://github.com/walletconnect/walletconnect-specs

bugbytesinc:
Is this backed by a hedera community HIP?

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,177 @@
...
+#### Request
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "method": "hedera_signAndExecuteTransaction",
+  "params": {
+    "transaction": {
+      "type": "CryptoTransfer",
```
bugbytesinc: "type" is information that is already contained within the transaction to be signed, it is redundant here and also any wallet would need to confirm that the encoded "bytes" indeed represent the the same type of transaction.  I'd suggest it be removed.

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,177 @@
...
+### Parameters
+
+    1. `Object` - signMessage parameters
+      1.1 `message` : `string` - base64 encoded string of arbitrary data
+
+### Returns
+
+    1. `Object` - signMessage result
+      1.1 `signature` : `string` - base64 encoded string of signed message bytes array
+
```
bugbytesinc, line 243: Perhaps a protobuf encoded SigMap would be a better choice here because it also contains necessary information required by the network for fusing mutliple signatures required for a transaction.  Additionally the controller can return multiple signatures if required.

docs/blockchain-rpc/hedera-rpc.md
hgraphql: the original idea was that there'd be some handling that could happen before if we know the type of transaction, though maybe that wouldn't have much utility?

hgraphql:
> Is this backed by a hedera community HIP?

it is not currently in a HIP, but I think that's a great suggestion. I'll get started on one

docs/blockchain-rpc/hedera-rpc.md
hgraphql: okay, I'll dive into this. thanks!

bugbytesinc:
> > Is this backed by a hedera community HIP?
> 
> it is not currently in a HIP, but I think that's a great suggestion. I'll get started on one

Started one: https://github.com/hashgraph/hedera-improvement-proposal/discussions/819

docs/blockchain-rpc/hedera-rpc.md
gregscullard: There may be an advantage to having this field for languages such as GoLang which aren't as "helpful" when it comes to converting bytes back to a transaction object, GoLang doesn't have a generic way of doing it with protobuf, you essentially have to try/catch each possible transaction type until it works from memory.
Such an attribute would enable a switch/case in GoLang which would be more efficient.

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,177 @@
...
+\*2. Transactions generated with the Hedera SDK can be converted to bytes with the [`toBytes` method](https://github.com/hashgraph/hedera-sdk-js/blob/50d7f76ffd1fd797a029192fab08b035720998b9/src/transaction/Transaction.js#L1015).
+
+### Returns
+
+    1. `Object` - Result of transaction submission to Hedera network
+      1.1. `response` : `Object` - Response from Hedera network from transaction submission
+        1.1.1. `nodeId` : `String` - The Hedera node the transaction was submitted to
+        1.1.2. `transactionHash` : `String` - The hash of the transaction
+        1.1.3. `transactionId` : `String` - Tranasction ID, which included the payer Account ID and the consensus timestamp
```
gregscullard: TransactionIDs famously have different formats in different places with combinations of "@.-", should we specify what the format is here ? Maybe it's `transactionId.toString()` from the SDK ?

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,177 @@
...
+      1.2 `receipt` : `TransactionReceipt Object` - Details of the completed transaction (*1)
```
gregscullard: Is this a serialised object in base64 ? JSON conversion of the receipt in SDKs isn't currently complete/compatible across all SDKs and therefore not preferred. An issue was raised with the SDK team to resolve.
The issue is now closed: https://github.com/hashgraph/hedera-sdk-js/issues/1855 and https://github.com/hashgraph/hedera-sdk-go/issues/797
(Should check that all SDKs yield the same JSON for a receipt)

docs/blockchain-rpc/hedera-rpc.md
hgraphql: okay, thanks @gregscullard 

I'll work though the differences with both the Go SDK and JS to flush out any gotchas

docs/blockchain-rpc/hedera-rpc.md
hgraphql: yep, we should be very specific with this. in fact looks like this description is wrong, as it references the consensus timestamp vs. the valid start timestamp

we can put a link to https://docs.hedera.com/hedera/sdks-and-apis/sdks/transactions/transaction-id

docs/blockchain-rpc/hedera-rpc.md
hgraphql: This PR will change based on [HIP-820](https://github.com/hashgraph/hedera-improvement-proposal/pull/820)

Will wait to update when it is in Review / Approved state and we've had time to build demo apps

docs/blockchain-rpc/hedera-rpc.md
bugbytesinc: Are you saying `one of` is not supported when deserializing protobuf in GO?

glitch-txs:
@huck-hgraph @bugbytesinc can we merge?

hgraphql:
> @huck-hgraph @bugbytesinc can we merge?

Hi @glitch-txs I'll be managing open PR's.

We have a meeting tomorrow to finalize our spec, and then I'll make the necessary updates to be ready to merge by the end of the day tomorrow.

hgraphql:
> > @huck-hgraph @bugbytesinc can we merge?
> 
> Hi @glitch-txs I'll be managing open PR's.
> 
> We have a meeting tomorrow to finalize our spec, and then I'll make the necessary updates to be ready to merge by the end of the day tomorrow.

@glitch-txs Here's our update - I've made the updates to this PR and will collect final feedback from stakeholders tomorrow

hgraphql:
hi @glitch-txs this is ready to be merged

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+- [`hedera_signMessage`](#hedera_signmessage)
+- [`hedera_getNodeAddresses`](#hedera_getnodeaddresses)
+
+## hedera\_signAndExecuteTransaction
+
+The `hedera_signAndExecuteTransaction` method is a generic method for executing a transaction on the Hedera network. A dApp can begin by constructing a transaction with one of the Hedera SDKs or by constructing the raw protobuf messages and may select one or more consensus nodes that are authorized to execute the transaction.
+
+The dApp then constructs a list of valid transaction bytes that differ only in the node account id and serializes the list, for example by using the `toBytes()` method of an SDK. Finally, the dApp base64 encodes the resulting bytes. This final base64 encoded string of bytes is sent as a method param titled `transactionList` to the wallet.
+
+Wallets and SDKs must take special care to verify that each transaction in the list differs only in the node that is authorized to submit the transaction and does NOT differ in intent before submitting to an end user for approval and ultimately signing.
```
bumblefudge:
```suggestion
Wallets and SDKs must take special care to deserialize the list of transactions and validate that each transaction in the list differs only in the node authorized to receive the transaction and does NOT differ in intent before submitting to an end user for approval and ultimately signing.
```

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+### Returns
+
+    1. `Object` - Result of transaction submission to Hedera network
+      1.1. `nodeId` : `String` - The Hedera node the transaction was submitted to
+      1.2. `transactionHash` : `String` - The hash of the transaction
+      1.3. `transactionId` : `String` - Transaction ID, which includes the payer account id and the valid start timestamp
+
+### Error
+
+In certain conditions, the Hedera network with return a response that signifies a pre-processing validation error, for example, when the transaction has expired. In these cases, wallets will return an error with the following format:
```
bumblefudge:
```suggestion
In certain conditions, the Hedera network will return a response that signifies a pre-processing validation error, for example, when the transaction has expired. In these cases, wallets will return an error with the following format:
```

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+
+    1. `Object` - Result of transaction submission to Hedera network
+      1.1. `code` : 9000 - the reserved Wallet Connect error code for unknown or errors not related to the Wallet Connect protocol
+      1.2. `message` : `String` - A human readable string describing the nature of the failure
```
bumblefudge, line 61: Is this human-readable string totally arbitrary/free-form? Is a blank message ok, or is the [string that corresponds to the applicable enum code](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/response_code.proto#L32) preferable as a fallback when no more custom or specific message is available?  I only mention it because familiarity with the error-code mapping might be a lot to ask of a dapp developer whose first interaction with the hashiverse is adding hedera wallet support to a dapp in "evm-rpc mode only" 😅

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+      1.1. `transactionHash` : `String` - The hash of the transaction
+      1.1. `transactionId` : `String` - Transaction ID, which includes the payer account id and the valid start timestamp
+
+### Error
+
+In certain conditions, the Hedera network with return a response that signifies a pre-processing validation error, for example, when the transaction has expired. In these cases, wallets will return an error with the following format:
+
+    1. `Object` - Result of transaction submission to Hedera network
+      1.1. `code` : 9000 - the reserved Wallet Connect error code for unknown or errors not related to the Wallet Connect protocol
+      1.1. `message` : `String` - A human readable string describing the nature of the failure
```
bumblefudge, line 166: same thought as above

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+## hedera\_getNodeAddresses
+
+While constructing a transaction for transmission to a controller, a dApp needs to choose which Hedera Network node shall receive the transaction prior to signing (this is a requirement of the Hedera API Protocol). While a dApp can easily obtain a list of potential Hedera Nodes, a controller may not have an all-inclusive list nor a path to the node’s gRPC endpoint. The hedera_getNodeAddresses method allows a dApp to request a list of node wallet addresses known to the controller. The controller should only include nodes in this list that it is willing and able to submit transactions to at the time of the request.
+
+### Returns
+
+    1. `Object` - signMessage result
+      1.1 `nodes` : `Array` - an array of strings <shard>.<realm>.<num>-<optional-checksum> format identifying an Hedera Address
```
bumblefudge:
```suggestion
      1.1 `nodes` : `Array` - an array of strings in trimmed `<shard>.<realm>.<num>-<optional-checksum>` format each identifying a node by its Hedera Address
```
nits: is the checksum truly, fully optional? can one member in the array have the checksum and the others not? will it be accepted without a checksum but with the `-` divider?  i'm assuming these nits aren't worth picking, but just in case the answer is 'no' to any of those, maybe change the word `in trimmed` to `in [validated](https://docs.hedera.com/hedera/core-concepts/accounts#account-id) and trimmed ` or otherwise link to the important nits 😄

bumblefudge (commented):
LGTM, nothing blocking here

docs/blockchain-rpc/hedera-rpc.md
hgraphql: great callout

In practice this will be similar to 
```javascript
transaction ${props.transactionId.toString()} failed precheck with status ${props.status.toString()};
```

This is from the javascript SDK here - https://github.com/hashgraph/hedera-sdk-js/blob/c78512b1d43eedf1d8bf2926a5b7ed3368fc39d1/src/PrecheckStatusError.js#L48C1-L49C1

So, the `props.status.toString()` piece should give the string that corresponds to error code - https://github.com/hashgraph/hedera-sdk-js/blob/c78512b1d43eedf1d8bf2926a5b7ed3368fc39d1/src/Status.js#L42

I'd probably recommend not setting hard requirements on the format of this string, though definitely open to that if that's helpful

docs/blockchain-rpc/hedera-rpc.md
hgraphql: the checksum is actually optional, here's the reference - https://hips.hedera.com/hip/hip-15#backwards-compatibility

though...the divider is only applicable with the checksum :)

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+## hedera\_getNodeAddresses
+
+While constructing a transaction for transmission to a controller, a dApp needs to choose which Hedera Network node shall receive the transaction prior to signing (this is a requirement of the Hedera API Protocol). While a dApp can easily obtain a list of potential Hedera Nodes, a controller may not have an all-inclusive list nor a path to the node’s gRPC endpoint. The hedera_getNodeAddresses method allows a dApp to request a list of node wallet addresses known to the controller. The controller should only include nodes in this list that it is willing and able to submit transactions to at the time of the request.
+
+### Returns
+
+    1. `Object` - signMessage result
+      1.1 `nodes` : `Array` - an array of strings in trimmed `<shard>.<realm>.<num>-<optional-checksum>` format each identifying a node by its Hedera Address
```
hgraphql:
```suggestion
      1.1 `nodes` : `Array` - an array of strings in `<shard>.<realm>.<num><-optional-checksum>` format, each identifying a node by its Hedera Address
```

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+
+The dApp then constructs a list of valid transaction bytes that differ only in the node account id and serializes the list, for example by using the `toBytes()` method of an SDK. Finally, the dApp base64 encodes the resulting bytes. This final base64 encoded string of bytes is sent as a method param titled `transactionList` to the wallet.
+
+Wallets and SDKs must take special care to deserialize the list of transactions and validate that each transaction in the list differs only in the node authorized to receive the transaction and does NOT differ in intent before submitting to an end user for approval and ultimately signing.
+
+
+### Parameters
+
+    1. `Object` - signAndExecuteTransaction parameters
+      1.1. `signerAccountId` : `String` - Hedera account id in the format `<network>:<shard>.<realm>.<num>-<optional-checksum>`
```
hgraphql:
```suggestion  
        1.1. `signerAccountId` : `String` - Hedera account id in the format `<network>:<shard>.<realm>.<num><-optional-checksum>`
    ```

File: docs/blockchain-rpc/hedera-rpc.md [hunk 12]
```diff
@@ -0,0 +1,302 @@
...
+}
+```
+## hedera\_signTransaction
+
+The `hedera_signTransaction` signs a `TransactionBody` and returns a `SignatureMap` to the caller.
+
+### Parameters
+
+    1. `Object` - signTransaction parameters
+      1.1 `signerAccountId` : `String` - hedera account id in the format `<network>:<shard>.<realm>.<num>-<optional-checksum>`
```
hgraphql:
```suggestion
      1.1 `signerAccountId` : `String` - hedera account id in the format `<network>:<shard>.<realm>.<num><-optional-checksum>`
```

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+```
+## hedera\_signAndExecuteQuery
+
+This method provides functionality to perform a query on a Hedera consensus node. Many Queries against consensus nodes have a transaction fee [Learn more about Queries](https://docs.hedera.com/hedera/sdks-and-apis/sdks/queries)
+Most request that do not change network state can be performed against a [Mirror Node](https://docs.hedera.com/hedera/core-concepts/mirror-nodes) and for many use cases calling a mirror node endpoint is the recommended approach.
+
+### Parameters
+
+    1. `Object` - signMessage parameters
+      1.1 `signerAccountId` : `String` - Hedera account id in the format `<network>:<shard>.<realm>.<num>-<optional-checksum>`
```
hgraphql:
```suggestion
      1.1 `signerAccountId` : `String` - Hedera account id in the format `<network>:<shard>.<realm>.<num><-optional-checksum>`
```

docs/blockchain-rpc/hedera-rpc.md
```diff
@@ -0,0 +1,302 @@
...
+
+```javascript
+
+<Ed25519 or ECDSA Key>.sign("\x19Hedera Signed Message:\n" + len(message) + message)
+```
+
+### Parameters
+
+    1. `Object` - signMessage parameters
+      1.1 `signerAccountId` : `String` - hedera account id in the format `<network>:<shard>.<realm>.<num>-<optional-checksum>`
```
hgraphql:
```suggestion
      1.1 `signerAccountId` : `String` - hedera account id in the format `<network>:<shard>.<realm>.<num><-optional-checksum>`
```

glitch-txs (approved):

PR #122, opened 2023-08-14
feat: update TTLs

chris13524:
Update notify message TTLs as per discussion today.

- JWT and message TTLs are equal
- How to handle race condition between JWT and message TTLs
- All responses (subscription response, subscription update response, subscription delete response, notify receipt) have 30 day TTLs to make sure the other side is able to receive them
- Message TTL remains 30 days to ensure the client receives it
- Subscription delete request TTL increased to 30 days to ensure notify server receives it (subscription expires after 30 days anyway), even if device submits it optimistically
- Subscription request and subscription update request TTLs reduced to 30 seconds to avoid unexpected changes with a subscription being created or updated days after the client submitted it

### DIFF
--- a/docs/specs/clients/notify/notify-authentication.md
+++ b/docs/specs/clients/notify/notify-authentication.md
@@ -11,7 +11,7 @@ In this document we will describe the authentication payloads for different meth
 - Notify Subscription Delete
 - Notify Subscription Delete Response
 
-All of the above authentication payloads are did-jwts and share the following claims:
+All of the above authentication payloads are DID JWTs and share the following claims:
 
 - act - description of action intent. Must be equal to specific value defined in each payload
 - iat - timestamp when JWT was issued
--- a/docs/specs/clients/notify/notify-authentication.md
+++ b/docs/specs/clients/notify/notify-authentication.md
@@ -26,7 +26,7 @@ JWT TTLs must be equal to the message TTLs they are sent on.
 
 If a client receives a JWT that is expired, it must ignore it as if it was never received. This is to ensure there are no race conditions between message TTL and JWT TTL.
 
-Another potential way to avoid the race condition is for the sender to set the message TTL to slightly less than the JWT TTL, however this is not ideal (and must not be done) as message TTL is a releative time from when the relay receives the message and not absolute like the JWT expiration is.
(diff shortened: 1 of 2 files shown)


llbartekll (approved):

pedrouid (approved):
👍

TalhaAli00 (approved):

pedrouid (approved):
👍

docs/specs/clients/notify/notify-authentication.md
```diff
@@ -11,27 +11,35 @@ In this document we will describe the authentication payloads for different meth
 - Notify Subscription Delete
 - Notify Subscription Delete Response
 
-All of the above authentication payloads will share the following claims:
+All of the above authentication payloads are did-jwts and share the following claims:
```
Cali93:
```suggestion
All of the above authentication payloads are DID JWTs and share the following claims:
```

docs/specs/clients/notify/rpc-methods.md
```diff
@@ -88,13 +86,14 @@ Used to publish a notification message to a peer through notify topic. Response
...
 | TTL     | 2592000  |
 | Tag     | 4003     |
-
 ```
 
 ### wc_notifyDelete
 
 Used to inform the peer to close and delete a notify subscription through notify topic. The reason field should be a human-readable message defined by the SDK consumer to be shown on the peer's side.
 
+Note: If the Notify Server is offline when the delete request is sent, the relay will retain the message the mailbox. The TTL is set such that when the Notify Server comes back online the delete request will be received. The message expiration is equal to the expiration of the subscription itself. Notify clients should assume deletion is successful as long as the relay ACKs the message. The delete response is for informational purposes and does not need to be received.
```
Cali93:
```suggestion
Note: If the Notify Server is offline when the delete request is sent, the relay will retain the message in the mailbox. The TTL is set such that when Notify Server comes back online, the delete request will be received. The message expiration is equal to the expiration of the subscription itself. Notify clients should assume deletion is successful as long as the relay ACKs the message. The delete response is for informational purposes and does not need to be received.
```

Cali93 (approved):
LGTM

PR #125, opened 2023-08-14
introduce relay client session identifier

pedrouid:
Introducing relay session lifecycle conditions and identifier to re-use the Relay Auth JWT subject which previously was previously assigned to a random nonce

### DIFF
--- a/docs/specs/clients/core/relay/relay-client-auth.md
+++ b/docs/specs/clients/core/relay/relay-client-auth.md
@@ -32,9 +32,23 @@ Client authentication is described with a did-jwt using the following claims:
 
 `iss` - did:key of client id key. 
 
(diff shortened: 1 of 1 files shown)


docs/specs/clients/core/relay/relay-client-auth.md
```diff
@@ -32,9 +32,23 @@ Client authentication is described with a did-jwt using the following claims:
 
 `iss` - did:key of client id key. 
 
-`aud` - relay server URL
+`aud` - relay server url
```
chris13524: URL is an acronym so should be caps

```suggestion
`aud` - relay server URL
```

jakubuid (approved):

PR #130, opened 2023-08-16
specs: change registration for Notify

Elyniss:


Cali93 (commented):
`register` makes totally sense.
I just have some questions because we couple it with `enableSync`, wouldn't we face some edge cases ?
For example, a user registers and decides to not enable sync but wants to enable it later on. Would it be ok to call `register` again ?

Elyniss:
>register makes totally sense.
I just have some questions because we couple it with enableSync, wouldn't we face some edge cases ?
For example, a user registers and decides to not enable sync but wants to enable it later on. Would it be ok to call register again ?

@Cali93 
Calling `register` again with `enableSync` param set to `true` should do the trick.

Cali93 (approved):

docs/specs/clients/notify/client-api.md
```diff
@@ -44,13 +43,16 @@ abstract class Client {
...
-  public abstract enableSync({
+  // registers a blockchain account with an identity key if not yet registered on this client
+  // if enableSync is true then register sync keys
+  // returns the public identity key.
+  public abstract register(params: {
         account: string, 
-        onSign: (message: string) => Cacao.Signature
-  }): Promise<void>;
+        onSign: (message: string) => Cacao.Signature,
+        enableSync: boolean
```
Talhaali00: Should we just enable this by default?

TalhaAli00 (approved):

llbartekll (approved):

docs/specs/clients/notify/client-api.md
Elyniss: I'm personally in favour of enabling Sync by default, and disallowing using Notify without Sync

pedrouid: Agreed. It should always be enabled. No need for this flag

Elyniss: Looks like we are all in agreement here. Flag removed

pedrouid (approved):
👍

devceline (approved):
LGTM

flypaper0 (approved):

PR #146, opened 2023-09-27
Update client-sdk-api.md

llbartekll:


docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -66,6 +66,9 @@ abstract class Client {
     onSign: (message: string) => Promise<Cacao.Signature>
   }): Promise<string>;
 
+  // deletes all subscriptions, decouples client from the registered account and pauses watchSubscription requests
```
chris13524: I don't think deleting all subscriptions makes sense to do here: https://walletconnect.slack.com/archives/C044SKFKELR/p1695827130071699?thread_ts=1695826302.572449&cid=C044SKFKELR

chris13524 (changes requested):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -66,6 +66,9 @@ abstract class Client {
...
+  public abstract logOut(): Promise<Void>;
```
Elyniss: How about we add two methods?
mute(account: string) (old logOut())
unmute(account: string)

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -66,6 +66,9 @@ abstract class Client {
     onSign: (message: string) => Promise<Cacao.Signature>
   }): Promise<string>;
 
+  // Unsubscribes from all subscription topics related to the most recently registered account, decouples the client from the registered account, and pauses watchSubscription requests.
```
devceline: What does `decouples the client from the registered account` mean

llbartekll: for instance service that is periodically triggering watchSubscription request is no longer aware of the account
probably depends of the implementation

devceline: Shouldn't it also delete the actual; identity key

Elyniss: The unregister is good when we want this device to forget about Notify API whatsoever. 

Clients are account agnostic. Why should we track most recently registered account? What is missing in he method is `account`. We had a phase in Chat API when we wanted to do similar thing. In the end we decided to fallback to always specifying the account for simplicity.


On the other hand,
I still think we should go mute/unmute route. Maybe not in this PR but in near future. Which would alternatively solve it.

llbartekll: >Shouldn't it also delete the actual; identity key

why? user may want to use this account again later

devceline: Then calling it `unregister` doesn't make sense

Elyniss: Then why would they unregister?

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -66,6 +67,13 @@ abstract class Client {
     onSign: (message: string) => Promise<Cacao.Signature>
   }): Promise<string>;
 
+  // exposed IdentityClient.unregister method
+  // unregister identity key for given account. Removes aforementioned identity key from local storage. Unregisters device from Push Server
+  // removes any logic to perodically call `wc_notifyWatchSubscriptions`
```
chris13524:
```suggestion
  // "Logs out" this client from any notifications for the specified account. This involves:
  // - Stops periodically calling `wc_notifyWatchSubscriptions` and unsubscribes from all related relay topics
  // - Unregisters the client from Push Server
  // - Unregisters the Identity Key and removes it from local storage
```

chris13524 (commented):
Can we cleanup the wording a bit?

flypaper0:
Also need to include `account` param into `getActiveSubscriptions()`

devceline (approved):
LGTM

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -66,6 +67,14 @@ abstract class Client {
     onSign: (message: string) => Promise<Cacao.Signature>
   }): Promise<string>;
 
+  // "Logs out" this client from any notifications for the specified account. This involves:
+  // - Stops periodically calling `wc_notifyWatchSubscriptions` and unsubscribes from all related relay topics
+  // - Unregisters the client from Push Server
+  // - Unregisters the Identity Key and removes it from local storage
```
chris13524, line 73:
```suggestion
  // - Unregisters the Identity Key and removes it from local storage
  // This is not to be confused with `deleteSubscription()` method
```

chris13524 (approved):

flypaper0 (approved):

Elyniss (approved):

PR #147, opened 2023-09-28
[SPECS]Decrypted web3wallet notifications

jakubuid:


docs/specs/meta-clients/web3wallet/sdk-api.md
```diff
@@ -71,7 +71,7 @@ class Web3Wallet {
   public abstract getPendingAuthRequests(): Promise<Record<number, PendingRequest>>;
   
   // register device token for Push server (BOTH)
-  public abstract registerDeviceToken(token: string): Promise<void>;
+  public abstract registerDeviceToken(token: string, isDecrypted: boolean?): Promise<void>;
```
chris13524: I think we should flip this language actually for clarity. Notifications were always un-encrypted, now they are becoming encrypted. We are registering the device token to receive encrypted notifications, not decrypted ones.

```suggestion
  public abstract registerDeviceToken(token: string, enableEncrypted: boolean?): Promise<void>;
```

docs/specs/servers/push/spec.md
```diff
@@ -61,14 +61,16 @@ POST <PUSH_SERVER_URL>/clients
...
+    "token": <DEVICE_TOKEN>,
+		"is_decrypted": <BOOLEAN> //optional
 }
 ```
 
 - `CLIENT_ID`: The Client's ID from the Relay pairing.
 - `TYPE`: The push service to use e.g. APNS, FCM. This will be validated against the supported types on the Push
   Server's side.
 - `DEVICE_TOKEN`: The device's token for the push service e.g. FCM, APNS.
+- `IS_DECRYPTED`: Defines when to send non-encrypted or encrypted content(Encrypted_Sign, Encrypted_Auth)
```
chris13524, line 73:
```suggestion
- `IS_DECRYPTED`: Defines when to send non-encrypted (false) or encrypted (true) content(Encrypted_Sign, Encrypted_Auth)
```

docs/specs/servers/push/spec.md
```diff
@@ -61,14 +61,16 @@ POST <PUSH_SERVER_URL>/clients
 {
     "client_id": <CLIENT_ID>,
     "type": <TYPE>,
-    "token": <DEVICE_TOKEN>
+    "token": <DEVICE_TOKEN>,
+		"is_decrypted": <BOOLEAN> //optional
```
chris13524, line 65:
```suggestion
    "token": <DEVICE_TOKEN>,
    "is_decrypted": <BOOLEAN> //optional
```

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
```diff
@@ -0,0 +1,31 @@
...
+
+    //Notification
+	{
+        "title": string,
+        "body": string,
+        "icon": string,
+        "url": string,
+        "type": string,
+    }
+```
```
chris13524:
```suggestion
```typescript
abstract class DecryptedNotification {
    // decrypt encrypted request
    public decryptMessage(params: { topic: string, encryptedMessage: string }): Promise<Message>;

    // get notification metadata for a given request
    public getNotificationMetadata(params: { topic: string, message: Message }): Promise<Notification>;
}

type Message = {
    id: string,
    jsonrpc: string, 
    method: string, 
    params: string, //Sign(proposal, request) params, Auth request params, Notify params
}

type Notification = {
    title: string,
    body: string,
    icon: string,
    url: string,
    type: string,
}
```
```

File: docs/specs/meta-clients/web3wallet/sdk-api.md (hunk 1 above)
Comment (jakubuid) [reply docs/specs/meta-clients/web3wallet/sdk-api.md]: Yeah from the server perspective, it makes more sense. I'm fine with changing

File: docs/specs/meta-clients/web3wallet/decrypted-norifications.md [hunk 5]
```diff
@@ -0,0 +1,27 @@
+# Decrypted Notification
+
+The below class introduces the general way of handling decprytped notifications received by a wallet. The implementation details differs on each platform.
```
Elyniss:
```suggestion
The below class introduces the general way of handling decrypted notifications received by a wallet. The implementation details differs on each platform.
```

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
```diff
@@ -0,0 +1,27 @@
...
+
+    // get notification metadata for a given request
+    public getNotificationMetadata(params: { topic: string, message: Message }): Promise<Notification>;
+}
+
+type Message = {
+    id: string,
+    jsonrpc: string, 
+    method: string, 
+    params: string, //Sign(proposal, request) params, Auth request params, Notify params
```
Elyniss:
```suggestion
    params: string,
```
Instead of loosely specifying what methods will be support I would add a dedicated section that enumerates every tag for each protocol

docs/specs/servers/push/spec.md
```diff
@@ -58,17 +58,28 @@ OK, push-server v0.1.0
...
 }
 ```
 
-- `CLIENT_ID`: The Client's ID from the Relay pairing.
-- `TYPE`: The push service to use e.g. APNS, FCM. This will be validated against the supported types on the Push
-  Server's side.
-- `DEVICE_TOKEN`: The device's token for the push service e.g. FCM, APNS.
+- `always_raw` - enables always sending of raw notifications
+  - When set to `true`, the push notification will be `topic` and `message` from Send Notifications below.
+  - When set to `false` or not present, the legacy behavior of conditionally sending (depending on `payload.flags`) encrypted or cleartext push notifications from `payload.blob` is used. The notification will contain `topic`, `flags`, and `blob` fields.
```
Elyniss:
```suggestion
- `always_raw` - forces sending raw notifications
  - When set to `true`, the push notification will be `topic` and `message` as per [Send Notifications](#send-notification) section.
  - When set to `false` or not present, the legacy behavior of conditionally sending (depending on `payload.flags`) encrypted or cleartext push notifications from `payload.blob` is used. The notification will contain `topic`, `flags`, and `blob` fields.
```

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +91,51 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
 
 ### Send Notification
 
+This API is called by the relay.
```
Elyniss:
```suggestion
This endpoint is called by the relay.
```

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +91,51 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+
+    // See below
+    flags: number,
+
+    // See below
+    blob: string,
+  }
+```
+
+- `flags`: an integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
```
Elyniss:
```suggestion
- `flags`: An integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
```

chris13524 (approved):

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +91,51 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
-    Notify     = 1 << 4
-    ```
-  - `blob`: If the payload is encrypted this is just the encrypted blob, when this is cleartext it should be base64 encoded. The blob should follow the format from of the [Notify Message](https://specs.walletconnect.com/2.0/specs/clients/notify/data-structures#notify-message)
+  // Topic of the message. This is used by the SDKs to decrypt encrypted payloads on the client side
+  topic: string,
+
+  // The relay message.
+  message: string,
+
+  // ==== Legacy Fields ====
```
chris13524: Migration milestones:
- Relay and Push Servers adopt this new spec (in no particular order)
- Wallet clients migrate to fully decrypted notifications
- Once all clients migrated, Push Servers remove Legacy Fields and default `always_raw: true`
- Once Push Servers default to `always_raw: true`, Legacy Fields are removed from relay

If we wish to remove the Legacy Fields from here, while still supporting clients that do not support decrypted notifications, then we will need to add additional specs that move the cleartext generation into Push Server. This will require passing `tag` from relay to Push Server (btw Webhooks 2.0 passes `tag`).

Context: https://walletconnect.slack.com/archives/C04DB2EAHE3/p1696257599856499?thread_ts=1695146414.042929&cid=C04DB2EAHE3

docs/specs/meta-clients/web3wallet/sdk-api.md
```diff
@@ -71,7 +71,7 @@ class Web3Wallet {
   public abstract getPendingAuthRequests(): Promise<Record<number, PendingRequest>>;
   
   // register device token for Push server (BOTH)
-  public abstract registerDeviceToken(token: string): Promise<void>;
+  public abstract registerDeviceToken(token: string, enableAlwaysDecrypted: boolean?): Promise<void>;
```
Talhaali00: since this is an addition, it doesn't need to be nullable. The flag will be missing only on older clients

docs/specs/servers/push/spec.md
```diff
@@ -58,17 +58,28 @@ OK, push-server v0.1.0
...
+  client_id: string,
+
+  // The push service to use.
+  type: "apns" | "apns-sandbox" | "fcm" | "noop",
+
+  // The device's token for the push service.
+  token: string,
+
+  // See below
+  always_raw?: boolean,
```
Talhaali00, line 76: since `false` and `not present` are the same, why not make this non-nullable and default to true?

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
jakubuid: Clients will use `method` to know how to parse the params. Maybe removing this comment would help?

docs/specs/servers/push/spec.md
jakubuid: Defaulting to true would mean that all clients support decrypted notifications which is not true. Clients to support this feature would have to set this flag to true on the SDK level and also set up a background service

docs/specs/meta-clients/web3wallet/sdk-api.md
jakubuid: @chris13524 do you have any objections to making this non-nullable on the server side?

arein: Default is `true` so only Rainbow/Robinhood set this to false until they upgraded their servers?

chris13524: `always_null` in the push server API needs to be nullable because it needs to support clients who have not updated the SDK. `null` is interpreted as `false`

Talha is talking about in the client SDK, which I think should be non-nullable. Or do we even need the parameter and just make it `true`?

chris13524: I agree with @arein, `enableAlwaysDecrypted` should default to `true` if not specified

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
Elyniss: Want I meant is that we should have documented somewhere which tags could be recevied, namely: 
- Sign: `SESSION_PROPOSE(1100)`, `SESSION_REQUEST*(1108)`
- Auth: `AUTH_REQUEST(3000)`
- Notify: `NOTIFY_MESSAGE(4002)`

chris13524: All of these messages would be received by the client. The client would filter them out via `method`

Elyniss: What I mean is that we had this specified in the technical brief. Hence, I think it would also be beneficent to have all those tags listed in the specs as well. If they are not documented how clients would know which methods to filter?

docs/specs/meta-clients/web3wallet/sdk-api.md
jakubuid: But there's a problem with defaulting `enableAlwaysDecrypted` to `true` because it means that we cut off the legacy users when they upgrade their SDK.
The upgrade path for wallets is to set this flag to true **AND** set up a background service
If they only set a flag without a background service, they would stop receiving legacy notifications

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
jakubuid: Gotcha, adding tags

docs/specs/meta-clients/web3wallet/sdk-api.md
chris13524: Yeah we'd have to mark it as a breaking version then. I also think there's going to be a longer drag on this, with the possibility that people choose not to do it. So I think it should default to `false` and I'm also going to update my migration plan above to enable push server to support non-decrypted notifications long-term.

docs/specs/servers/push/spec.md
chris13524: > we will need to add additional specs that move the cleartext generation into Push Server

I'll work on this

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
chris13524: @jakubuid the way you worded this sounds like echo-server will only push certain tags to client, but it will push all and clients need to filter. I reworded in [5066871](https://github.com/WalletConnect/walletconnect-specs/pull/147/commits/50668714bf25925a7e45fbfc17e682b79e2a63cd)

docs/specs/meta-clients/web3wallet/sdk-api.md
chris13524: I updated the default to `false` in [264bd16](https://github.com/WalletConnect/walletconnect-specs/pull/147/commits/264bd16db8c681cb3598b1deacac26756385510a)

docs/specs/servers/push/spec.md
chris13524: Updated specs to show that Push Server will now generate the cleartext message: [d83fc21](https://github.com/WalletConnect/walletconnect-specs/pull/147/commits/d83fc21af042c411ba08dfa16c6526324314d0f8)

Legacy Fields are now fully deprecated, and updated Push Server versions should not depend on them. Once Push Servers update, we can remove these fields from the relay. We can now handle the migration (forced or not) to decrypted notifications separately.

docs/specs/meta-clients/web3wallet/decrypted-norifications.md
jakubuid: Hmm, so that's why I wanted to include a tag in the notification payload, to know what message to filter out before decryption. Now clients have to first decrypt to know a method and decide whether to display a given notification.

If we had a tag in the payload, clients would filter out notifications by a tag and decrypt only those that are needed. It seems more efficient. Wdyt?

chris13524: Makes sense. `tag` added as custom notification data here: https://github.com/WalletConnect/walletconnect-specs/pull/147/commits/d83fc21af042c411ba08dfa16c6526324314d0f8

chris13524 (approved):

docs/specs/servers/push/spec.md
jakubuid: @chris13524 Does it break legacy users? Or does it keep them receiving simple/unencrypted notifications?

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +94,53 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+
 ```
 POST <PUSH_SERVER_URL>/clients/<CLIENT_ID>
+```
+
+- `CLIENT_ID` - This is used to map to the device token
+
+Headers:
+- `X-Ed25519-Timestamp` - timestamp of the signature
+- `X-Ed25519-Signature` - hex encoded ed25519 signature of `format!("{}.{}.{}", timestamp, body.len(), body)` where `body` is the request body
```
jakubuid, line 107: @chris13524 why do we need this signature and which signature is it?

docs/specs/servers/push/spec.md
chris13524: @jakubuid legacy users will continue receiving simple notifications, unless always_raw is set.

Here are updated migration steps:
- Relay and Push Servers adopt this new spec (in no particular order)
- Once both servers migrated, Relay and Push Servers remove Deprecated Fields

Optional:
- All wallet clients migrate to fully decrypted notifications
- Push Servers force always_raw: true, and simple/unencrypted notifications support removed

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +94,53 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
-    Auth       = 1 << 2
-    Chat       = 1 << 3
-    Notify     = 1 << 4
-    ```
-  - `blob`: If the payload is encrypted this is just the encrypted blob, when this is cleartext it should be base64 encoded. The blob should follow the format from of the [Notify Message](https://specs.walletconnect.com/2.0/specs/clients/notify/data-structures#notify-message)
+  // Topic of the message. This is used by the SDKs to decrypt encrypted payloads on the client side
+  topic: string,
+
+  // Tag of the message.
+  tag: string,
```
chris13524:
```suggestion
  tag: number,
```

docs/specs/servers/push/spec.md
chris13524: This is the relay's signature that signs the webhook so that echo server knows the content is legitimate. This was already implemented, but undocumented, so I added it here.

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +94,53 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+  payload: {
+    // Same as `topic` above
+    topic: string,
+
+    // See below
+    flags: number,
+
+    // See below
+    blob: string,
+  }
```
jakubuid, line 135: @chris13524 can you confirm that new fields that are added are inside the data objects of FCM
https://firebase.google.com/docs/cloud-messaging/concept-options#data_messages

chris13524: This is for the webhook from relay to push server. The notification content is specified under the Register Client section: https://github.com/WalletConnect/walletconnect-specs/pull/147/files/1a9ad6005447d6d7d6458824579ecc103f4af825#diff-25490e5dda4bcd46802fca4131439d45ab85cf408683b76c7604735dfe656edcR80

docs/specs/meta-clients/web3wallet/decrypted-notifications.md
```diff
@@ -0,0 +1,33 @@
...
+    method: string, 
+    params: string,
+}
+
+type Notification = {
+    title: string,
+    body: string,
+    icon: string,
+    url: string,
+    type: string,
```
chris13524:
```suggestion
    type: string,
    title: string,
    body: string,
    icon: string,
    url: string,
```

Elyniss (approved):

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +94,53 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+
+- `flags`: An integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
+  ```
+  Encrypted  = 1 << 0
+  Sign       = 1 << 1
+  Auth       = 1 << 2
+  Chat       = 1 << 3
+  Notify     = 1 << 4
+  ```
+- `blob`: If the flags specify encrypted, then this is just the relay message. If not encrypted, this is cleartext and should follow the format from of the [Notify Message](https://specs.walletconnect.com/2.0/specs/clients/notify/data-structures#notify-message)
```
chris13524:
```suggestion
- `blob`: If the flags specify encrypted, then this is just the relay message. If not encrypted, this is base64 encoded cleartext and should follow the format from of the [Notify Message](https://specs.walletconnect.com/2.0/specs/clients/notify/data-structures#notify-message)
```

docs/specs/servers/push/spec.md
```diff
@@ -58,17 +58,31 @@ OK, push-server v0.1.0
...
 }
 ```
 
-- `CLIENT_ID`: The Client's ID from the Relay pairing.
-- `TYPE`: The push service to use e.g. APNS, FCM. This will be validated against the supported types on the Push
-  Server's side.
-- `DEVICE_TOKEN`: The device's token for the push service e.g. FCM, APNS.
+- `always_raw` - forces sending raw notifications, defaults to `false`
+  - When set to `true`, the push notification will contain, as custom data, `topic`, `tag`, and `message` from the request payload in [Send Notifications](#send-notification) section.
+  - When set to `false`, cleartext notifications will be sent for certain messages based on `tag`, but all other messages follow the same behavior as `always_raw=true` and will be sent encrypted. Cleartext messages:
```
chris13524:
```suggestion
  - When set to `false`, cleartext notifications will be sent for certain messages based on `tag`, but all other messages will contain, as custom data, `payload.topic` and `payload.blob` from the request payload in [Send Notifications](#send-notification) section. Cleartext messages:
```

chris13524 (commented):
cc @geekbrother @jakubuid for some fixes

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +94,53 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+    blob: string,
+  }
+```
+
+- `flags`: An integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
+  ```
+  Encrypted  = 1 << 0
+  Sign       = 1 << 1
+  Auth       = 1 << 2
+  Chat       = 1 << 3
```
llbartekll, line 143: Do we need to include chat?

llbartekll (changes requested):
as discussed, move DecryptedNotification to dedicated protocol api specs, one per sdk.

remove Notification type

docs/specs/meta-clients/web3wallet/decrypted-notifications.md
```diff
@@ -0,0 +1,33 @@
...
+
+type Message = {
+    id: string,
+    jsonrpc: string, 
+    method: string, 
+    params: string,
+}
+
+type Notification = {
+    type: string,
```
chris13524: Latest Notify Message spec also has a notification ID which is to be used for deduplication

```suggestion
    id: string,
    type: string,
```

llbartekll: yeah but it should not be in the Push spec

jakubuid: Changed in: https://github.com/WalletConnect/walletconnect-specs/pull/147/commits/1a95259130de3d9904c3451aba260dc20f8c1538

jakubuid:
> as discussed, move DecryptedNotification to dedicated protocol api specs, one per sdk.
> 
> remove Notification type

Specs for Decrypted Notifications have been divided into Notify and Web3Wallet(Sign, Auth) (https://github.com/WalletConnect/walletconnect-specs/commit/1a95259130de3d9904c3451aba260dc20f8c1538)

docs/specs/meta-clients/web3wallet/decrypted-notifications.md
```diff
@@ -0,0 +1,32 @@
...
+```typescript
+abstract class DecryptedNotification {
+    // decrypt encrypted request
+    public decryptMessage(params: { topic: string, encryptedMessage: string }): Promise<Message>;
+
+    // get notification metadata for a given request
+    public getMetadata(params: { topic: string }): Promise<Metadata>;
+}
+
+type Message = {
```
jakubuid: Change for RPC request

docs/specs/clients/notify/decrypted-notifications.md
```diff
@@ -0,0 +1,21 @@
+# Decrypted Notifications
+
```
jakubuid, line 2: Add reference in push section

docs/specs/clients/notify/decrypted-notifications.md
```diff
@@ -0,0 +1,21 @@
...
+The below class introduces the general way of handling notify notifications received by a wallet. This should be handled in the background service where the implementation details differs on each platform.
+
+```typescript
+abstract class DecryptedNotification {
+    // decrypt encrypted request
+    public decryptMessage(params: { topic: string, encryptedMessage: string }): Promise<NotifyMessage>;
+}
+
+type NotifyMessage = {
+    id: string,
```
jakubuid, line 12: Revert type

docs/specs/meta-clients/web3wallet/decrypted-notifications.md
```diff
@@ -0,0 +1,32 @@
...
+    method: string,
+    params: string,
+}
+
+type Metadata = {
+    name: string,
+    description: string,
+    url: string,
+    icons: string[],
+    url: string,
```
jakubuid: Typo

docs/specs/meta-clients/web3wallet/sdk-api.md
```diff
@@ -71,7 +71,7 @@ class Web3Wallet {
   public abstract getPendingAuthRequests(): Promise<Record<number, PendingRequest>>;
   
   // register device token for Push server (BOTH)
-  public abstract registerDeviceToken(token: string): Promise<void>;
+  public abstract registerDeviceToken(token: string, enableAlwaysDecrypted?: boolean = false): Promise<void>;
```
jakubuid: enableAlwaysDecrypted - > maybe a different name?
@llbartekll

docs/specs/servers/push/spec.md
```diff
@@ -80,27 +94,53 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+  // Topic of the message. This is used by the SDKs to decrypt encrypted payloads on the client side
+  topic: string,
+
+  // Tag of the message.
+  tag: number,
+
+  // The relay message.
+  message: string,
+
+  // ==== Deprecated fields ====
```
jakubuid: Maybe Legacy fields? A better name?

chris13524:
```suggestion
  // ==== Legacy fields ====
```

docs/specs/servers/push/background-service
```diff
@@ -0,0 +1,6 @@
+### Background Service
+
+Each platform is responsible for handling decrypted notifications in background services received by a wallet where the implementation details differns on each platforms. The background service decides based on tag the type of notification to handle.
```
chris13524:
```suggestion
Each platform is responsible for handling decrypted notifications in background services received by a wallet where the implementation details differ on each platforms. The background service decides based on tag the type of notification to handle.
```

docs/specs/meta-clients/web3wallet/sdk-api.md
llbartekll: i would call them encrypted as they are e2ee.

llbartekll: also im not sure "always" adds any value in this naming 🤔

jakubuid: Yeah "always" is redundant but still not sure about encrypted vs decrypted. have to sleep on it :)

llbartekll: enableEncrypted? wdyt @jakubuid

chris13524: How about:
- `enableDetailed`
- `enableAdvanced`

?

jakubuid: `enableAdvanced` or `enableEncrypted` are the best choices IMO

jakubuid: `enableEncrypted` let it be

ganchoradkov (approved):
🥇

llbartekll (approved):

geekbrother (approved):

PR #148, opened 2023-09-28
chore: document relay <-> push interaction, improve decrypted notifications specs

chris13524:
- Added specs on the relay <-> push server interaction.
- Changed from `is_decrypted` to `always_raw` and `enableAlwaysEncrypted` as I think this makes more sense.

I think we should flip this language actually for clarity. Notifications were always un-encrypted, now they are becoming encrypted (so it makes sense to say "is encrypted" not "is decrypted"). We are registering the device token to receive encrypted notifications, not decrypted ones.


docs/specs/servers/push/spec.md
```diff
@@ -84,27 +93,42 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+  }
+```
+
+- `flags`: an integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
+  ```
+  Encrypted  = 1 << 0
+  Sign       = 1 << 1
+  Auth       = 1 << 2
+  Chat       = 1 << 3
+  Notify     = 1 << 4
```
jakubuid, line 139:
```suggestion
  Encrypted  = 1 << 0
  Sign       = 1 << 1
  Auth       = 1 << 2
  Chat       = 1 << 3
  Notify     = 1 << 4
  Encrypted_Sign = 1100 / 1108 (session_propose / session_request)
  Encrypted_Auth = 3000 (auth_request) 
```

SDK would need these two new flags to know how to parse a blob

chris13524: > SDK would need these two new flags to know how to parse a blob

I don't think so because if the push client sets `always_raw: true` then the raw encrypted message would always be sent and it would always need to be decrypted.

chris13524: Pushed a commit adding the details of what I think the PNs would contain: 7d89303

jakubuid: My point is that even though the `always_raw` flag is set to true, for SDK it would be beneficial to know the type of the request. It would be needed to know the type to which SDK should parse JSON

It has nothing to do with decryption

By sending a request tag as a notification flag we could easily achieve this

jakubuid (approved):

docs/specs/servers/push/spec.md
chris13524: (We agreed the JSON-RPC `method` can be used to parse the JSON)

docs/specs/servers/push/spec.md
```diff
@@ -82,29 +91,46 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
+  // Message ID for deduplication purposes
+  // Note: When `payload.raw` is set, `sha264(payload.raw)` should be used as the Message ID instead
+  id: string,
+
+  payload: {
+    // Topic of the message. This is used by the SDKs to decrypt encrypted payloads on the client side
+    topic: string,
+
+    // Raw message sent on the topic
+    raw: string,
```
chris13524:
```suggestion
    // Message sent on the topic
    message: string,
```

docs/specs/servers/push/spec.md
```diff
@@ -82,29 +91,46 @@ DELETE <PUSH_SERVER_URL>/clients/<CLIENT_ID>
...
-    Sign       = 1 << 1
-    Auth       = 1 << 2
-    Chat       = 1 << 3
-    Notify     = 1 << 4
-    Encrypted_Sign = 1100 / 1108 (session_propose / session_request)
-    Encrypted_Auth = 3000 (auth_request)  
-    ```
-  - `blob`: If the payload is encrypted this is just the encrypted blob, when this is cleartext it should be base64 encoded. The blob should follow the format from of the [Notify Message](https://specs.walletconnect.com/2.0/specs/clients/notify/data-structures#notify-message)
+  // Message ID for deduplication purposes
+  // Note: When `payload.raw` is set, `sha264(payload.raw)` should be used as the Message ID instead
```
chris13524:
```suggestion
  // Note: When `payload.message` is set, `sha264(payload.message)` should be used as the Message ID instead
```

docs/specs/servers/push/spec.md
```diff
@@ -58,19 +58,28 @@ OK, push-server v0.1.0
...
 }
 ```
 
-- `CLIENT_ID`: The Client's ID from the Relay pairing.
-- `TYPE`: The push service to use e.g. APNS, FCM. This will be validated against the supported types on the Push
-  Server's side.
-- `DEVICE_TOKEN`: The device's token for the push service e.g. FCM, APNS.
-- `IS_DECRYPTED`: Defines when to send non-encrypted or encrypted content(Encrypted_Sign, Encrypted_Auth)
+- `always_raw` - enables always sending of raw notifications
+  - When this is set to `true`, the pushed notification will always be the value from `payload.raw` from Send Notifications below. The notification will contain `topic` and `message` fields.
```
chris13524:
```suggestion
  - When this is set to `true`, the pushed notification will always be the value from `payload.message` from Send Notifications below. The notification will contain `topic` and `message` fields.
```

PR #150, opened 2023-10-03
feat: notifyWatchSubscriptions refresh updates

chris13524:
Clarifies behavior when notifyWatchSubscriptions is called multiple times. Currently, Notify Server will create a new watcher for each call, but the desired behavior involves updating the watcher instead.


Elyniss (approved):

devceline (approved):
LGTM

chris13524:
Implemented in https://github.com/WalletConnect/notify-server/pull/113

PR #158, opened 2023-10-07
[Verify API] Remove verifyUrl

pedrouid:


chris13524 (approved):

pedrouid:
Disabled at the API level... no need to change the protocol spec

PR #160, opened 2023-10-09
feat: nullable url and icon

chris13524:
This makes the `url` and `icon` fields nullable. Clients should have fallback behavior e.g. no click behavior and no icon.

### DIFF
--- a/docs/specs/clients/notify/data-structures.md
+++ b/docs/specs/clients/notify/data-structures.md
@@ -26,7 +26,7 @@
   title: string,
   // Long messages used in the body of the notification
   body: string,
-  // Image URL used to display with the notification
+  // Image URL used to display with the notification. If null, the app's icon from Notify Config is used instead
   icon: string | null,
   // Redirect URL for call-to-action related to notification
   url: string | null,
--- a/docs/specs/clients/notify/data-structures.md
+++ b/docs/specs/clients/notify/data-structures.md
@@ -18,13 +18,18 @@
 
 ## Notify Message
 
-```jsonc
+```typescript
 {
-  "title": string, // short message used in the title of the notification
-  "body": string, // long messages ued in the body of the notification
-  "icon": string, // image URL used to display with the notification
-  "url": string, // redirect URL for call-to-action related to notification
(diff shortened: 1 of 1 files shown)


docs/specs/clients/notify/data-structures.md
```diff
@@ -18,13 +18,18 @@
...
-  "url": string, // redirect URL for call-to-action related to notification
-  "type": string, // notification type which matches the scope of notify subscription
+  // Notification type which matches the scope of notify subscription
+  type: string,
+  // Short message used in the title of the notification
+  title: string,
+  // Long messages used in the body of the notification
+  body: string,
+  // Image URL used to display with the notification
+  icon: string | null,
```
devceline, line 30:
```suggestion
  icon?: string,
```

I believe [this](https://specs.walletconnect.com/2.0/specs/clients/notify/client-sdk-api) is the format we use

docs/specs/clients/notify/data-structures.md
```diff
@@ -18,13 +18,18 @@
...
+  // Redirect URL for call-to-action related to notification
+  url: string | null,
```
devceline, line 32:
```suggestion
  url?: string,
```

devceline (commented):
After this gets merged we should also add this in the docs repo

docs/specs/clients/notify/data-structures.md
chris13524: For dev-facing things, yes, but here this is an internal data structure so the field should be nullable rather than optional.

docs/specs/clients/notify/data-structures.md
chris13524: Same as https://github.com/WalletConnect/walletconnect-specs/pull/160#discussion_r1352594014

devceline (approved):
LGTM

flypaper0 (approved):

chris13524:
Added detail if no icon was specified, to use the dapp's icon from Notify Config instead. Context: https://walletconnect.slack.com/archives/C044SKFKELR/p1697465670021389?thread_ts=1697447230.241749&cid=C044SKFKELR

PR #161, opened 2023-10-09
feat: messaging refactor

chris13524:


Cali93 (approved):

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
+type Body = [{
+  // Idempotency key and ID to track status
+  notificationId?: string | null,
+  notification: {
+    // The notification type ID found in [Notify Config](../../clients/notify/notify-config.md)
+    type: string,
+    title: string,
+    body: string,
+    icon?: string | null,
+    url?: string | null,
```
Elyniss: Shouldn't be nullable

chris13524: @Elyniss it should because this is the Notify API endpoint. `null` will be translated to `""` by the Notify Server

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
-{
-  "notification": {
-    "title": string,
-    "body": string,
-    "icon": string,
-    "url": string,
-    "type": string,
+type Body = [{
+  // Idempotency key and ID to track status
+  notificationId?: string | null,
```
geekbrother, line 16: Do we really need to give the ability to provide the ID for a new notification? For me, server-generated IDs look more consistent, and does the client need this ability? cc @chris13524

chris13524: The primary purpose of this ID is actually for idempotency.

We can have an idempotencyKey field instead, and require the client to read the response in order to call the status endpoint. But I figured it's a slightly nicer API for the caller to be able to specify their own IDs in order to avoid storing the response to their database if they already have notification IDs on their end.

In my approach this notificationId would be dapp-facing only, unique key being separate per-dapp, and the pk of the database would still be a server-generated UUID. Happy with either approach really, but we need idempotency.

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
 ```typescript
-Account[]
-``` 
-
-## Subscribe Topic
+type Response = [{
+  account: AccountId,
+  // Notification types subscribed to
+  scope: string[],
+}]
```
chris13524: Make changing the response body easier in the future w/o introducing a new endpoint version:

```suggestion
type Response = {
  subscribers: {
    account: AccountId,
    // Notification types subscribed to
    scope: string[],
  }[],
}
```

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
-{
-  "id": string, // webhook ID
+type Response = WebhookResponse[]
+
+type WebhookResponse = Webhook & {
+  // Webhook ID
+  id: string,
+  // Shared secret for authenticating Webhook Requests
+  secret: string,
 }
```
chris13524: Make changing the response body easier in the future w/o introducing a new endpoint version:

```suggestion
type Response = {
  webhooks: (Webhook & {
    // Webhook ID
    id: string,
    // Shared secret for authenticating Webhook Requests
    secret: string,
  })[],
}
```

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
+```typescript
+// CAIP-10 account ID
+type AccountId = string
+```
 
 ```typescript
-string
+type Response = {
+  // IDs for the sent notifications returned in the same order as the body. If a `notificationId` wasn't specified in the request, one will be generated.
+  notifications: string[],
```
devceline, line 42: With the queuing behavior now, are consumers of this API expected to query all the messages they sent until they either get `published` or some of the error statuses? 

Not sure if I missed something here

chris13524: > are consumers of this API expected to query all the messages they sent until they either get published or some of the error statuses?

Yes, but only if they want to know the status. However I suppose this is polling so would be good to have a webhook option for this as well.

devceline: yes - I think a webhook option would make sense. Polling would be absurd for all the different notification ids

docs/specs/servers/notify/notify-server-api.md
devceline, line 16: since notification id is on the level of the entire request, where one notification could include many users,, how would we account to different success statuses amongst users?

Say notif A was sent to users 1,2,3. 1 had successful message delivery, 2 was not subscribed at all, 3 was not subscribed to the notification type.

What would the status of that notification be? how would users of this API who to resend to?

chris13524: See the endpoint below for your answer. The status isn't per notification, but per account for a notification.

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
+    type: string,
+    title: string,
+    body: string,
+    icon?: string | null,
+    url?: string | null,
   },
-  "accounts": Account[],
-}
+  // The accounts to send this notification to
+  accounts: AccountId[],
```
chris13524:
```suggestion
  // The accounts to send this notification to. If null or undefined, all accounts subscribed to the notification type will be sent the notification.
  accounts?: AccountId[] | null,
```

docs/specs/servers/notify/notify-server-api.md
```diff
@@ -2,183 +2,244 @@
...
 
 ## Subscribers 
 
-Get the list of all accounts currently subscribed to this dapp.
+Get the list of all accounts currently subscribed to this app.
 
-`GET /subscribers`
-
-Response:
+`GET /v1/<project-id>/subscribers`
```
chris13524, line 176:
```suggestion
`GET /v1/<project-id>/subscribers`

Query params:
- Only return subscribers that are subscribed to 1 or more notification types: `?scope=<notification-type>[,<notification-type>]`
```

PR #166, opened 2023-10-20
Make icon and url not nullable again

Elyniss:
We have to revert this as this is a breaking change on older clients. Since Notify should no longer have too many braking changes we cannot make this change anymore

### DIFF
--- a/docs/specs/clients/notify/data-structures.md
+++ b/docs/specs/clients/notify/data-structures.md
@@ -28,7 +28,7 @@
   body: string,
   // Image URL used to display with the notification. If empty, the app's icon from Notify Config is used instead
   icon: string,
-  // Redirect URL for call-to-action related to notification.
+  // Redirect URL for call-to-action related to notification. If empty, do not redirect
   url: string,
 }
 ```
--- a/docs/specs/clients/notify/data-structures.md
+++ b/docs/specs/clients/notify/data-structures.md
@@ -26,10 +26,10 @@
   title: string,
   // Long messages used in the body of the notification
   body: string,
-  // Image URL used to display with the notification. If null, the app's icon from Notify Config is used instead
-  icon: string | null,
-  // Redirect URL for call-to-action related to notification
-  url: string | null,
+  // Image URL used to display with the notification. If empty, the app's icon from Notify Config is used instead
+  icon: string,
+  // Redirect URL for call-to-action related to notification.
+  url: string,
 }
 ```
 


devceline (approved):
LGTM

docs/specs/clients/notify/data-structures.md
```diff
@@ -26,10 +26,10 @@
   title: string,
   // Long messages used in the body of the notification
   body: string,
-  // Image URL used to display with the notification. If null, the app's icon from Notify Config is used instead
-  icon: string | null,
-  // Redirect URL for call-to-action related to notification
-  url: string | null,
+  // Image URL used to display with the notification. If empty, the app's icon from Notify Config is used instead
+  icon: string,
```
chris13524, line 30: To make it clear to expect this to be an empty string. The other variables will be guaranteed by Notify Server not to be empty.

```suggestion
  icon: string | "",
```

docs/specs/clients/notify/data-structures.md
```diff
@@ -26,10 +26,10 @@
...
+  // Redirect URL for call-to-action related to notification.
+  url: string,
```
chris13524, line 32:
```suggestion
  url: string | "",
```

docs/specs/clients/notify/data-structures.md
devceline: `string | ""` looks weird. "" is a string, the comment "If empty.." makes it clear it can be empty

docs/specs/clients/notify/data-structures.md
devceline: maybe we can mention "Can be empty" in the comment above like the `icon`

docs/specs/clients/notify/data-structures.md
Elyniss: Isn't this confusing?

Elyniss: @devceline thoughts?

devceline: my thoughts are above

docs/specs/clients/notify/data-structures.md
Elyniss:
```suggestion
  url: string, // if value is empty this means no redirection
```

docs/specs/clients/notify/data-structures.md
Elyniss: My UI didn't update, lol

chris13524 (approved):

flypaper0 (approved):

PR #171, opened 2023-10-31
feat: Sign SDK Wallet Authentication

ganchoradkov:
- Added Wallet Authentication ([CAIP-222](https://github.com/ChainAgnostic/CAIPs/pull/248)) to Sign SDK specs
- Removed Auth API specs


docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,24 @@ abstract class Client {
     reason: Reason;
   }): Promise<void>;
 
+  // ---------- Wallet Authentication Methods ----------------------------------------------- //
+
+  // for proposer to create an authenticated pairing
```
llbartekll: authenticated pairing?
we have never covered that concept 😅

ganchoradkov: fixed the wording

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,24 @@ abstract class Client {
     reason: Reason;
   }): Promise<void>;
 
+  // ---------- Wallet Authentication Methods ----------------------------------------------- //
+
+  // for proposer to request wallet authentication
+  public abstract authenticate(params: {
```
llbartekll, line 80: will it work?

llbartekll: it use to be RequestParams
```
    public let domain: String
    public let chainId: String
    public let nonce: String
    public let aud: String
    public let nbf: String?
    public let exp: String?
    public let statement: String?
    public let requestId: String?
    public let resources: [String]?
```

llbartekll: probably we can simplify 🤔

ganchoradkov: we can, but saw that I missed some required parameters so for the specs I'm listing now all available

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
...
+    "metadata": Metadata
+  },
+  "payloadParams": PayloadParams,
+  "expiry": number // optional
+}
+
+| IRN     |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Tag     | 3000     |
```
llbartekll:
```suggestion
| Tag     | 1116     |
```

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
...
+[{
+  "h": CacaoHeader,
+  "p": CacaoPayload,
+  "s": CacaoSignature
+}]
+
+| IRN     |          |
+| ------- | -------- |
+| TTL     | 86400    |
+| Tag     | 3001     |
```
llbartekll:
```suggestion
| Tag     | 1117     |
```

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,33 @@ abstract class Client {
...
+    exp?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+    pairingTopic?: string;
+}): Promise<{ uri: string }>;
+
+
+   // respond to wallet authentication request
+  public abstract respondSessionAuthenticated(params: RespondParams, iss: string): Promise<boolean>;
```
jakubuid: We probably don't boolean as a return type

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,33 @@ abstract class Client {
     reason: Reason;
   }): Promise<void>;
 
+  // ---------- Wallet Authentication Methods ----------------------------------------------- //
+
+  // for proposer to request wallet authentication
+  public abstract authenticate(params: {
```
jakubuid, line 80: Will ReCaps be added on the next PR/step?

docs/specs/clients/sign/data-structures.md
```diff
@@ -175,8 +177,10 @@ Verify Context is appended to Session Proposals and Session Requests to provide
 }
 ```
 
-## Namespace Config
+### Namespace Config
```
jakubuid, line 181: lol, we have it in specs but not implemented yet

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
...
+Parameters for request and response are compliant with [CAIP-74](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-74.md) (aka CACAO)
+
+- Success response is equivalent to authentication acceptance.
+- Error response is equivalent to authentication rejection.
+
+##### Expiry
+
+Param `Expiry` is an optional Unix timestamp. Sets the time until which the responder can respond to this request. If request is expired responder should respond with a specific error code.
+
+If this parameter is not specified, the request is considered indefinite.
```
jakubuid, line 379: I suggest setting a default value for Expiry, let's say 5mins instead of considering it indefinite when not specified. What's the value of having a request being indefinite?

ganchoradkov: good idea although I would say to increase the expiry as singinig with MPC wallet would take much longer and proposer can't know in advance. We can have it by default 1d or 7d? Wdyt?

docs/specs/clients/sign/client-api.md
ganchoradkov: Yes, this is the first step, after ReCaps standard & specs

docs/specs/clients/sign/rpc-methods.md
jakubuid: 1d sounds good

docs/specs/clients/sign/README.md
```diff
@@ -20,6 +20,28 @@ This swipe puts the dapp (usually in a browser window) on the background which m
...
+
+Sign API's wallet authentication is a one-click passwordless authentication provider for any website or application.
+
+### Goals
+
+A user can login or sign-in into any website or application with their blockchain account.
+
+An app does not need to maintain a persisted connection with the wallet once authenticated.
+
+A wallet can evaluate all auth parameters requested to format the message independently.
```
jakubuid, line 45: The message must be formatted in an exact precise way to avoid signature mismatches, right? So we'll expose formatMessage method for that to help devs. Having that, shouldn't we remove this part: "to format the message independently" since it might cause issues?

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,33 @@ abstract class Client {
...
+    exp?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+    pairingTopic?: string;
+}): Promise<{ uri: string }>;
+
+
+   // respond to wallet authentication request
+  public abstract respondSessionAuthenticated(params: RespondParams, iss: string): Promise<void>;
```
jakubuid, line 96:
```suggestion
  public abstract respondAuthenticatedSession(params: RespondParams, iss: string): Promise<void>;
```

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
...
+```jsonc
+// Success
+result: true
+
+// Failure
+error: {
+  "code": number,
+  "message": string
+}
+```
```
jakubuid, line 362: We already have definitions at the top of the file. I think it's redundant to have them twice

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
 | TTL     | 30       |
 | Tag     | 1115     |
 ```
+
+## Wallet Authentication RPC Methods
```
jakubuid, line 346: Why do we need a dedicated section for this? `wc_sessionAuthenticated` is becoming a part of the standard Sign features suite so it should be enough to add it within all other methods

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
...
+
+```jsonc
+// wc_sessionAuthenticated params
+{
+  "requester": {
+    "publicKey": string,
+    "metadata": Metadata
+  },
+  "payloadParams": PayloadParams,
+  "expiry": number // optional
```
jakubuid: It shouldn't be optional, because as we agreed above we don't want to have indefinite sessions

docs/specs/meta-clients/web3wallet/README.md
```diff
@@ -1,3 +1,3 @@
 # Web3Wallet SDK Overview
 
-The Web3Wallet SDK combines functionality of [Sign](../../clients/sign/README.md) and [Auth](../../clients/auth/README.md) into one offering by giving wallets a dedicated SDK that bundles both APIs together and only exposing the relevant methods
+The Web3Wallet SDK simplifies the functionality of [Sign](../../clients/sign/README.md) by giving wallets SDK that bundles both Sign & Wallet Authentication together and only exposing the relevant methods
```
jakubuid: We should add methods and events related to wallet authentication to Web3Wallet client API

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
...
 
+  // ---------- Wallet Authentication Methods ----------------------------------------------- //
+
+  // for proposer to request wallet authentication
+  public abstract sessionAuthenticate(params: {
+    chains: string[];
+    domain: string;
+    nonce: string;
+    aud: string;
+    type?: CacaoHeader["t"];
```
jakubuid, line 85: Why optional?

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
...
+    nbf?: string;
+    exp?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+    expiry?: number;
```
jakubuid, line 91: what's the difference between `expiry` and `exp`?

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
     reason: Reason;
   }): Promise<void>;
 
+  // ---------- Wallet Authentication Methods ----------------------------------------------- //
+
+  // for proposer to request wallet authentication
+  public abstract sessionAuthenticate(params: {
```
jakubuid: Seems like `"iat": string` is missing

docs/specs/clients/sign/rpc-methods.md
jakubuid: We agreed to remove `"expiry": number`

docs/specs/clients/sign/rpc-methods.md
```diff
@@ -338,3 +342,87 @@ true
...
+**Request**
+
+```jsonc
+// wc_sessionAuthenticated params
+{
+  "requester": {
+    "publicKey": string,
+    "metadata": Metadata
+  },
+  "payloadParams": PayloadParams,
```
jakubuid: Let's decide on one name, above we have `authPayloads` here `payloadParams`

docs/specs/clients/sign/session-authenticated-events.md
```diff
@@ -0,0 +1,134 @@
...
+                        "aud": "aud",
+                        "domain": "localhost",
+                        "version": "1",
+                        "nonce": "1",
+                        "iat": "2023-12-18T15:35:00.148Z",
+                        "resources": [
+                            "urn:recap:eyJhdHQiOnsiZWlwMTU1IjpbeyJyZXF1ZXN0L3BlcnNvbmFsX3NpZ24iOltdfSx7InJlcXVlc3QvZXRoX3NpZ25UeXBlZERhdGFfdjQiOltdfV19fQ=="
+                        ],
+                        "chainId": "eip155:1",
+                        "iss": "did:pkh:eip155:1:0x73bA24E3AA21546E0F4bC87d10bEE3d8bd637C79"
```
jakubuid:
```
                    "p": {
                        "type": "caip122",
                        "chains": [
                            "eip155:1",
                            "eip155:2"
                        ],
                        "aud": "aud",
                        "domain": "localhost",
                        "version": "1",
                        "nonce": "1",
                        "iat": "2023-12-18T15:35:00.148Z",
                        "resources": [
                            "urn:recap:eyJhdHQiOnsiZWlwMTU1IjpbeyJyZXF1ZXN0L3BlcnNvbmFsX3NpZ24iOltdfSx7InJlcXVlc3QvZXRoX3NpZ25UeXBlZERhdGFfdjQiOltdfV19fQ=="
                        ],
                        "chainId": "eip155:1",
                        "iss": "did:pkh:eip155:1:0x73bA24E3AA21546E0F4bC87d10bEE3d8bd637C79"
```

According to CAIP-222 there are no chains nor chainId since it can be gotten from iss
Am I missing sth?

docs/specs/clients/sign/session-authenticated-events.md
```diff
@@ -0,0 +1,134 @@
...
+         "type":"caip122",
+         "chains":[
+            "eip155:1",
+            "eip155:2"
+         ],
+         "aud":"aud",
+         "version":"1",
+         "domain":"https://walletconnect.com",
+         "nonce":"1",
+         "iat":"2023-12-18T15:31:28.071Z",
```
jakubuid, line 40: Shouldn't we have methods here?

docs/specs/clients/sign/client-api.md
jakubuid: Because we set type internally - it's optional in case people want to set it

docs/specs/clients/sign/client-api.md
ganchoradkov: `expiry` to be used for request ttl, `exp` for cacao expiry

docs/specs/clients/sign/data-structures.md
```diff
@@ -195,3 +199,160 @@ Namespace Config is set within the setNamespaceConfig method that is a part of t
...
+  id: number;
+  topic: string;
+  params: {
+    requester: Participant;
+    authPayload: PayloadParams;
+  };
+}
+```
+
+### Request Params
```
jakubuid, line 225: Let's highlight that these are client api params

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
...
+    chains: string[];
```
jakubuid, line 81: Only single namespace validation

docs/specs/clients/sign/client-api.md
jakubuid: @ganchoradkov can you reflect on it?

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
...
+    aud: string;
+    type?: CacaoHeader["t"];
+    nbf?: string;
+    exp?: string;
+    statement?: string;
+    requestId?: string;
+    resources?: string[];
+    expiry?: number;
+    pairingTopic?: string;
+    methods?: string[];
```
jakubuid, line 93: `methods` being optional allow authenticated session without ability to send requests. is it sth that we aim for?

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
...
+  }): Promise<{
+      uri: string,
+      response: Promise<{
+        auths?: Cacao[];
```
jakubuid, line 97: Why optional?

docs/specs/clients/sign/README.md
```diff
@@ -20,6 +20,28 @@ This swipe puts the dapp (usually in a browser window) on the background which m
...
+
+### Context
+
+Sign API's wallet authentication would be used for wallet users to login in a single step into existing websites or applications that would normally require an email and password login or a social login such as Facebook, Google, etc
+
+Sign API's wallet authentication is a one-click passwordless authentication provider for any website or application.
+
+### Goals
+
+A user can login or sign-in into any website or application with their blockchain account.
```
Talhaali00: to help with consistency since sign-in isn't used anywhere else

```suggestion
A user can login into any website or application with their blockchain account.
```

docs/specs/meta-clients/web3wallet/README.md
Talhaali00:
```suggestion
The Web3Wallet SDK simplifies the functionality of [Sign](../../clients/sign/README.md) by giving wallets an SDK that bundles both Sign & Wallet Authentication together and only exposing the relevant methods
```

TalhaAli00 (approved):

docs/specs/clients/sign/session-authenticated-events.md
```diff
@@ -0,0 +1,134 @@
...
+         "publicKey":"49006eb041860891a010ce23ba981b095499271d0df80a3a4b53f936fb57773e",
+         "metadata":{
+            "name":"App A (Proposer)",
+            "description":"Description of Proposer App run by client A",
+            "url":"https://walletconnect.com",
+            "icons":[
+               "https://avatars.githubusercontent.com/u/37784886"
+            ]
+         }
+      }
```
ganchoradkov, line 55: to add after `requester`
```
      expiryTimestamp: 1xxxxxx,
```

docs/specs/clients/sign/client-api.md
ganchoradkov: This has been discussed already but to have it written down, not specifying `methods` will perform single SIWE request without establishing authentication session

docs/specs/clients/sign/client-api.md
ganchoradkov: Dynamic switcher e.g. approving normal sign session doesn't produce `auths`. Having it option or as empty `[]` array imo is the same

docs/specs/clients/sign/client-api.md
llbartekll, line 91:
```suggestion
    ttl?: number; // ttl of a request, in seconds from now
```

what's default 3600?

llbartekll: > expiry to be used for request ttl

I think we should call it ttl and it's for calculating request expiry, not related to relay ttl, relay ttl can be always 5 minutes

ganchoradkov: 1h in seconds

llbartekll: ok but do we call it ttl or expiry?

ganchoradkov: `expiry`is more intuitive imo

docs/specs/clients/sign/README.md
```diff
@@ -20,6 +20,28 @@ This swipe puts the dapp (usually in a browser window) on the background which m
...
+
+### Description
+
+Sign API provides a simple and lean interface to enable applications to authenticate wallet users to login/signin into applications with their wallets by automatically signing an authentication message.
+
+### Context
+
+Sign API's wallet authentication would be used for wallet users to login in a single step into existing websites or applications that would normally require an email and password login or a social login such as Facebook, Google, etc
+
+Sign API's wallet authentication is a one-click passwordless authentication provider for any website or application.
```
llbartekll, line 37: it's copy/paste from auth api, sounds like siwe only

docs/specs/clients/sign/README.md
```diff
@@ -20,6 +20,28 @@ This swipe puts the dapp (usually in a browser window) on the background which m
...
+
+### Goals
+
+A user can login into any website or application with their blockchain account.
+
+An app does not need to maintain a persisted connection with the wallet once authenticated.
```
llbartekll, line 43: we could mention sessions here too

docs/specs/clients/sign/client-api.md
```diff
@@ -88,5 +132,10 @@ abstract class Client {
 
   // subscribe to session delete
   public abstract on("session_delete", (sessionDelete: { id: number, topic: string }) => {}): void;
+
+  // ---------- Wallet Authentication Events ----------------------------------------------- //
+
+  // wallets to subscribe to session authenticated
+  public abstract on("session_authenticate", (sessionAuthenticateRequest: SessionAuthenticateRequest) => {}): void;
```
llbartekll, line 139: this is SessionAuthenticateRequest and you don't have on response event, does mean that if app crashes you are loosing track of the response?

docs/specs/clients/sign/rpc-methods.md
llbartekll: 1h -> 3600s

docs/specs/clients/sign/client-api.md
pedrouid: Technically for the other sessions we have a method called `connect` 

Wouldn't it make more sense for this method to be called `authenticate`?

docs/specs/clients/sign/client-api.md
pedrouid: Yeah this is really confusing to have the same concept duplicated... TTL would be easier

docs/specs/clients/sign/client-api.md
pedrouid: Shouldn't that be cached internally with the SignClient? You can use the `getPendingAuthRequests`

docs/specs/clients/sign/rpc-methods.md
pedrouid: Agreed

docs/specs/clients/sign/session-authenticated-events.md
```diff
@@ -0,0 +1,130 @@
+import Table from '../../../components/Table';
+
+# Session Authenticate Events
```
pedrouid:
```suggestion
# Session Authenticated Events
```

docs/specs/clients/sign/wallet-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+         "request/eth_signTypedData_v4":[{
+            chains: ['eip155:1']
+         }],
+         "request/personal_sign":[{
+            chains: ['eip155:1']
+         }]
+      }
+   }
+}
+```
```
pedrouid, line 161: I think we need to add another example here where two chains are supported and add some language around the authentication

Consider the following scenarios when chains `eip155:1` and `eip155:2` are `requested`

Scenario A - the wallet only supports `eip155:1` then we simply receive a one CACAO with an array that includes only `eip155:1`  (This is already included in the example above)

Scenario B - the wallet supports both `eip155:1  and `eip155:2` then we receive two CACAOs with an array that includes both `eip155:1` and `eip155:2` in each CACAO (this means that we can guarantee the wallet is authenticated for both chains and we can send requests for both chains)

Scenario C -  the wallet supports both `eip155:1  and `eip155:2` BUT we only receive one CACAO with an array that includes both `eip155:1` and `eip155:2` (this means that we CANNOT guarantee the wallet is authenticated for both chains BUT we can send requests for both chains)

docs/specs/clients/sign/client-api.md
llbartekll:
```suggestion
  public abstract authenticate(params: {
```

docs/specs/clients/sign/data-structures.md
```diff
@@ -196,3 +200,162 @@ Namespace Config is set within the setNamespaceConfig method that is a part of t
...
+
+```typescript
+interface PayloadParams {
+  type: string; // same as Cacao Header type (t)
+  chains: string[];
+  domain: string;
+  aud: string;
+  version: string;
+  nonce: string;
+  iat: string;
```
jakubuid, line 262: If this is a part of the session authenticate request, we should remove `iat`. Wallet should set it.

jakubuid: @ganchoradkov @llbartekll

docs/specs/clients/sign/client-api.md
jakubuid: We should have "session_authenticate_response" for a dapp

docs/specs/clients/sign/wallet-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+
+B sends response on response topic.
+
+A receives response and validates signature.
+
+If signature is valid, then user is authenticated.
+
+## Capabilities
+
+Applications can request consent from users to perform actions on their behalf. These actions are included in the authentication request in the form of ReCaps.
```
bumblefudge:
```suggestion
Applications can request consent from users to perform actions on their behalf. These actions are included in the authentication request in the form of ReCaps, a compact encoding for authorization statements.
```

docs/specs/clients/sign/wallet-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+WalletConnect's implementation follows closely the [ERC-5573 specs](https://eips.ethereum.org/EIPS/eip-5573). ReCaps are also used as a form of handshake to establish Sign session.
```
bumblefudge:
```suggestion
WalletConnect's implementation follows closely the [ERC-5573 specs](https://eips.ethereum.org/EIPS/eip-5573). 
ReCaps are also used to include additional fine-grained authorization information along with the WCv2 handshake that establishes a Sign session.
```

docs/specs/clients/sign/wallet-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+- urn:recap:eyJhdHQiOnsiZWlwMTU1Ijp7InJlcXVlc3QvZXRoX3NpZ25UeXBlZERhdGFfdjQiOlt7fV0sInJlcXVlc3QvcGVyc29uYWxfc2lnbiI6W3t9XX19fQ==
+```
+
+The recap in the message
+
+```
+- urn:recap:eyJhdHQiOnsiZWlwMTU1Ijp7InJlcXVlc3QvZXRoX3NpZ25UeXBlZERhdGFfdjQiOlt7fV0sInJlcXVlc3QvcGVyc29uYWxfc2lnbiI6W3t9XX19fQ==
+```
+
+after removing the prefix (`urn:recap:`) and decoding would result in
```
bumblefudge:
```suggestion
after removing the URN prefix (`urn:recap:`), Base64-decoding the string would result in the JSON object:
```

docs/specs/clients/sign/wallet-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+I further authorize the stated URI to perform the following actions on my behalf: (1) 'request': 'eth_signTypedData_v4', 'personal_sign' for 'eip155'. (2) 'push': 'messages', 'notification' for 'eip155'. (3) 'receive': 'messages', 'notification' for 'eip155'.
+
+URI: https://example.com
+Version: 1
+Chain ID: 1
+Nonce: 1
+Issued At: 2024-02-19T09:29:21.394Z
+Resources:
+- urn:recap:eyJhdHQiOnsiZWlwMTU1Ijp7InJlcXVlc3QvZXRoX3NpZ25UeXBlZERhdGFfdjQiOlt7fV0sInJlcXVlc3QvcGVyc29uYWxfc2lnbiI6W3t9XX19fQ==
+- https://example.com
```
bumblefudge: having a bare "example.com" here is a little confusing, better to give a realistic example of what example might encode as a URI as an "additional authorization" not covered by the SIWE itself.
```suggestion
- https://example.com/storage/0x3613699A6c5D8BC97a08805876c8005543125F09
```

docs/specs/clients/sign/wallet-authentication.md
```diff
@@ -0,0 +1,161 @@
...
+- urn:recap:eyJhdHQiOnsiZWlwMTU1Ijp7InB1c2gvbWVzc2FnZXMiOlt7fV0sInB1c2gvbm90aWZpY2F0aW9uIjpbe31dfX19
+- urn:recap:eyJhdHQiOnsiZWlwMTU1Ijp7InJlY2VpdmUvbWVzc2FnZXMiOlt7fV0sInJlY2VpdmUvbm90aWZpY2F0aW9uIjpbe31dfX19
+```
+
+### Authenticated sessions
+
+As mentioned above, applications utilize ReCaps to request consent from users to establish authenticated session to be able to send additional requests for the approved methods & chains to the users' wallets
```
bumblefudge:
```suggestion
As mentioned above, applications utilize ReCaps to request additional permissions or capabilities from users, beyond what the basic capabilities (chains, accounts, methods) that get negotiated in a WC connection between application and wallet wallet.
```

docs/specs/clients/sign/wallet-authentication.md
bumblefudge: ^ Totally agree, this would clarify the difference between what a SIWE signature "proves" and what a ReCap "expresses" (and can maybe take on faith when combined with the SIWE signature), since they are both authorization statements.

ganchoradkov: Extra examples added 👍

docs/specs/clients/sign/session-authenticated-events.md
```diff
@@ -0,0 +1,130 @@
...
+         "type":"caip122",
+         "chains":[
+            "eip155:1",
+            "eip155:2"
+         ],
+         "aud":"aud",
+         "version":"1",
+         "domain":"https://walletconnect.com",
+         "nonce":"1",
+         "iat":"2023-12-18T15:31:28.071Z",
```
jakubuid, line 40: This shouldn't be here. `iat` is set by a wallet not by a dapp

docs/specs/clients/sign/client-api.md
```diff
@@ -74,6 +74,50 @@ abstract class Client {
...
   }): Promise<void>;
 
+  // ---------- Wallet Authentication Methods ----------------------------------------------- //
+
+  // for proposer to request wallet authentication
+  public abstract authenticate(params: {
+    chains: string[];
+    domain: string;
+    nonce: string;
+    aud: string;
```
llbartekll: are we changing for uri?

llbartekll (approved):

jakubuid (approved):

docs/specs/clients/sign/README.md
Cali93: I agree with Bartek, it'd be clearer if we mention sessions.

```suggestion
An app does not need to maintain a persisted connection with the wallet once authenticated thanks to a session guaranteeing ownership of the account.
```

PR #173, opened 2023-11-03
feat: Verify Server API spec NG

xDarksome:
Updates Verify Server API spec to support Native dApp attestation

### DIFF
--- a/docs/specs/servers/verify/verify-server-api.md
+++ b/docs/specs/servers/verify/verify-server-api.md
@@ -1,30 +1,169 @@
 # Verify Server API
 
-## Register Attestation
+## Generate Challenge (iOS only)
 
-Used to register a new attestation
+Generates a unique one-time challenge to be later used with other requests as a mean to prevent replay attacks.
 
-`POST /attestation`
+### Request
 
-Body:
+`POST /challenge`
+
+### Responses
+
+- `201` - Challenge generated
 
 ```jsonc
 {
-    "attestationId": string,
-    "origin": string
+    "id": string
 }
 ```
 
-## Resolve Attestation
+## Register Attestor (iOS only)
 
-Used to resolve an attestation
+Registers an instance of iOS dApp.
 
-`GET /attestation/:attestationId`
+["Verify the Attestation"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576643) step from Apple docs.
 
-Response:
+### Request
 
+`POST /attestor/:id`
+```jsonc
+{
+    "attestationObject": [number] 
+}
+```
+- `attestationObject` - byte array representing CBOR encoded ["Attestation Object"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576643)
+
+### Responses
+
+- `201` - Attestor registered
+- `400` - Invalid request body or `attestationObject` format
+- `401` - Failed to verify the `attestationObject` according to ["Verify the Attestation"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576643)
+- `403` - App ID is not registered on [WalletConnect Cloud](https://cloud.walletconnect.com)
+
+## Register Attestation
+
+Registers dApp attestation.
+
+### Web
+
+Uses our custom iframe Enclave solution.
+
+#### Request
+
+`POST /attestation/:id`
 ```jsonc
 {
     "origin": string
+}
+```
+- `origin` - Web dApp origin 
+
+#### Responses
+
+- `201` - Attestation registered
+- `400` - Invalid request body
+
+### Android
+
+Uses [Google Play Integrity API](https://developer.android.com/google/play/integrity/overview).
+
+["Decrypt and verify the integrity verdict"](https://developer.android.com/google/play/integrity/standard#decrypt-and) step from Android docs.
+
+#### Request
+
+`POST /attestation/:id`
+```jsonc
+{
+    "integrityToken": string
+}
+```
+- `integrityToken` - token provided to the Android dApp by [Integrity API](https://developer.android.com/google/play/integrity/standard#request-integrity)
+
+#### Responses
+
+- `201` - Attestation registered
+- `400` - Invalid request body or `integrityToken` format 
+- `401` - Failed to verify the `integrityToken` according to ["Decrypt and verify the integrity verdict"](https://developer.android.com/google/play/integrity/standard#decrypt-and)
+- `403` - Bundle ID is not registered on [WalletConnect Cloud](https://cloud.walletconnect.com)
+
+### iOS
+
+Uses [Apple DeviceCheck](https://developer.apple.com/documentation/devicecheck/).
+
+["Verify the Assertion"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576644) step from Apple docs.
+
+#### Request
+
+`POST /attestation/:id`
+```jsonc
+{   
+    "keyId": string, 
+    "assertion": [number]
+}
+```
+- `keyId` - key identifier used when registering the Attestor
+- `assertion` - byte array representing CBOR encoded ["Assertion Object"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576644)
+
+#### Responses
+
+- `201` - Attestation registered
+- `400` - Invalid request body or `assertion` format
+- `401` - Failed to authenticate the provided `keyId`
+
+## Resolve Attestation
+
+Resolves previously registered dApp attestation for a wallet. 
+
+### Web
+
+#### Request
+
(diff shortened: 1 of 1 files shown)


docs/specs/servers/verify/verify-server-api.md
```diff
@@ -1,30 +1,169 @@
...
+}
+```
+- `integrityToken` - token provided to the Android dApp by [Integrity API](https://developer.android.com/google/play/integrity/standard#request-integrity)
+
+#### Responses
+
+- `201` - Attestation registered
+- `400` - Invalid request body or `integrityToken` format 
+- `401` - Failed to verify the `integrityToken` according to ["Decrypt and verify the integrity verdict"](https://developer.android.com/google/play/integrity/standard#decrypt-and)
+- `403` - Bundle ID is not registered on [WalletConnect Cloud](https://cloud.walletconnect.com)
```
jakubuid, line 88: @Cali93 can you add specs for the Explorer endpoint, where dapp's metadata URL is returned based on the bundle_id?

docs/specs/servers/verify/verify-server-api.md
```diff
@@ -1,30 +1,169 @@
...
+}
+```
+- `attestationObject` - byte array representing CBOR encoded ["Attestation Object"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576643)
+
+### Responses
+
+- `201` - Attestor registered
+- `400` - Invalid request body or `attestationObject` format
+- `401` - Failed to verify the `attestationObject` according to ["Verify the Attestation"](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server#3576643)
+- `403` - App ID is not registered on [WalletConnect Cloud](https://cloud.walletconnect.com)
```
jakubuid, line 42: @Cali93 can you add specs for the Explorer endpoint, where dapp's metadata URL is returned based on the app_id?

I think when making app_id unique in Explorer, we don't have to send the project_id
@llbartekll can two apps on AppleStore have the same apple_id?

docs/specs/servers/verify/verify-server-api.md
```diff
@@ -1,30 +1,169 @@
...
+`GET /attestation/:id`
+
+#### Responses
+
+- `200` - Attestation exists
+    - `appId` - App ID of the iOS dApp that registered this attestation
+```jsonc
+{
+    "appId": string
 }
```
jakubuid, line 168: `GET /attestation/:id` should be one endpoint across all platforms (web, iOS, Android) because a wallet doesn't know if a given request comes from web, iOS or Android Dapp.

This assumption forces having one response:
{
    "origin": string // for native it'll be metadata URL returned from the Explorer
    "bindleId": string?, //optional - it's android bundle_id or iOS app_id 
    "isScam": boolean? //optional - for native it can be true when attestation exists but it indicates that dapp is not legit
}

xDarksome: Hm, so `Android` <-> `iOS` doesn't make sense, obviously. Valid combinations of dApp -> Wallet are:
- Web -> Anything? (is Web wallet a thing?)
- Android -> Android
- iOS -> iOS

So technically the responses should be:
- iOS - `(origin, isScam) | appId`
- Android - `(origin, isScam) |  bundleId`

We don't have scam checking for non-Web dApps.

Makes sense?

docs/specs/servers/verify/verify-server-api.md
```diff
@@ -1,30 +1,169 @@
 # Verify Server API
 
+## Generate Challenge (iOS only)
+
+Generates a unique one-time challenge to be later used with other requests as a mean to prevent replay attacks.
+
+### Request
+
+`POST /challenge`
```
llbartekll, line 9: why not GET?

docs/specs/servers/verify/verify-server-api.md
llbartekll: ofc not

docs/specs/servers/verify/verify-server-api.md
xDarksome: Actually, as we want to return a sum type maybe we should have those fields inside an object that explicitly defines the attestation source.
```json
{
  "web": { "origin": "..", "isScam": false }
  "android": { "bundleId": ".." }
  "ios": { "appId": ".." }
}
```
The response would contain only 1.
Android  expects `web` and `android`, iOS expects `web` and `ios`

Adding more platforms doesn't seem to break this approach. Because generally a platform cares only about a dApps from it's own platform + `web`

jakubuid:
Couple of steps that have to happen on VerifyServer for Android attestation:
- Get `requestPackageName` from the `requestDetails` object from the decoded Integrity Token - use it for the Explorer endpoint to check if whitelisted and to get metadata URL
- On the resolve call - get `requestHash` from the `requestDetails` object from the decoded Integrity Token - compare it with `attestation_id` sent by wallet, they must be the same
- in `appIntegrity` object from the decoded Integrity Token - `appRecognitionVerdict` must be `PLAY_RECOGNIZED`
- in `deviceIntegrity` object from the decoded Integrity Token - `deviceRecognitionVerdict` must be `MEETS_DEVICE_INTEGRITY`
- in `accountDetails` object from the decoded Integrity Token - `appLicensingVerdict ` must be `LICENSED `

cc @xDarksome 

Reference: https://developer.android.com/google/play/integrity/verdicts

docs/specs/servers/verify/verify-server-api.md
xDarksome: `GET` shouldn't generally be used for something that creates a resource on the backend. The main consideration is caching.

docs/specs/servers/verify/verify-server-api.md
llbartekll: for iOS we can use bundleId too but I am still not sure this approach makes sense

xDarksome: I expect the field set to evolve independently, so it should be more easy to comprehend the differences between platforms. But you are the one who's going to consume it, so it's your call. 
Technically, from the backend perspective I don't care whether it's a tagged or untagged union.

xDarksome:
@jakubuid sounds good

> On the resolve call - get requestHash from the requestDetails object from the decoded Integrity Token - compare it with attestation_id sent by wallet, they must be the same

I would rather do this check on `POST /attestation`, does it make any difference to you?

docs/specs/servers/verify/verify-server-api.md
jakubuid: Basically, all combinations are possible:
-  iOS wallet to iOS, web, Android dapp
- Android wallet iOS, web, Android dapp
- web wallet iOS, web, Android dapp 

A wallet shouldn't be aware of dapp's platform, it should only know if a given request is verified, unknown, scam, mismatched

Also, we can't change the existing GET /attestation:id response, because it'll break existing users
- `isScam` is optional so it's not a problem for native - we can use it anyway in the future
- `origin` will be the metadata URL from Explorer
- bundle_id - a field for Android bundle_id and iOS app_id, I cannot see issue having one field for both

llbartekll: Im still sceptical of the value of `bundle_id` 🤔

jakubuid: Yeah, I also don't see reasons for it now, we can skip it and add later

xDarksome: How `Android` <-> `iOS` is possible?

> A wallet shouldn't be aware of dapp's platform, it should only know if a given request is verified, unknown, scam, mismatched

This abstraction already leaks anyway, because native wallet needs to know about both `origin` and `bundleId` and `bundleId` may be null.
If you really want to abstract this away, then I should probably do it server side. And just return you a boolean or something.

> Also, we can't change the existing GET /attestation:id response, because it'll break existing users

Good point

Okay, if you guys want an object with optional fields, so be it.

xDarksome: Ah, so you don't actually need the `bundleId` but do need `origin`?

llbartekll: Let's give it a few more days 😅 I still feel like we haven't broken it yet

jakubuid:
> @jakubuid sounds good
> 
> > On the resolve call - get requestHash from the requestDetails object from the decoded Integrity Token - compare it with attestation_id sent by wallet, they must be the same
> 
> I would rather do this check on `POST /attestation`, does it make any difference to you?

@xDarksome 
Not really, just wanted to highlight the how Server should interpret verdicts. However, the result of the verdict should be consumed by a wallet on GET/attestation

docs/specs/servers/verify/verify-server-api.md
jakubuid: > Ah, so you don't actually need the bundleId but do need origin?

It seems so but we definitely need origin from Explorer

PR #177, opened 2023-11-28
Add isRegistered, getSiweMessage

devceline:
- Add `isRegistered` function
- Add `getSiweMessage` function
- Refactor `register` function

Matching docs PR: https://github.com/WalletConnect/walletconnect-docs/pull/1227/files

### DIFF
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -37,26 +37,26 @@ abstract class Client {
     account: string,
   }): Promise<Record<string, NotifySubscription>>;
 
-  // get all messages for a subscription
-  public abstract getMessageHistory(params: {
+  // get all notifications for a subscription
+  public abstract getNotificationHistory(params: {
     topic: string,
-  }): Promise<Record<number, NotifyMessageRecord>>
+  }): Promise<Record<number, NotifyNotificationRecord>>
 
   // delete active subscription
   public abstract deleteSubscription(params: {
     topic: string,
   }): Promise<void>;
   
-  // delete notify message
-  public abstract deleteNotifyMessage(params: {
+  // delete notification
+  public abstract deleteNotification(params: {
     id: number,
   }): Promise<void>;
   
   // decrypt notify subscription message
-  public abstract decryptMessage(params: {
+  public abstract decryptNotification(params: {
     topic: string,
     encryptedMessage: string,
-  }): Promise<NotifyMessage>;
+  }): Promise<NotifyNotification>;
   
   // Generates and returns SIWE message along with
   // Cacao Payload with identity key
@@ -103,8 +103,8 @@ abstract class Client {
   // for wallet to listen for notify subscription created
   public abstract on("notify_subscription", (result: NotifySubscription | Error) => {}): void;
   
-  //  for wallet to listen on notify messages
-  public abstract on("notify_message", (message: NotifyMessageRecord, metadata: Metadata) => {}): void;
+  //  for wallet to listen on notify notification
+  public abstract on("notify_notification", (message: NotifyNotificationRecord, metadata: Metadata) => {}): void;
   
   // for wallet to listen for result of notify subscription update
   public abstract on("notify_update", (result: NotifySubscription | Error) => {}): void;
--- a/docs/specs/clients/notify/data-structures.md
+++ b/docs/specs/clients/notify/data-structures.md
@@ -16,7 +16,7 @@
 }
 ```
 
-## Notify Message
+## Notify Notification
 
 ```typescript
 {
@@ -33,14 +33,14 @@
 }
 ```
 
-## Notify Message Record
+## Notify Notification Record
 
 ```jsonc
 {
   "id": string,
   "topic": string,
   "publishedAt": Int64,
-  "message": NotifyMessage
+  "message": NotifyNotification
 }
 ```
 


docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -56,17 +58,32 @@ abstract class Client {
     encryptedMessage: string,
   }): Promise<NotifyMessage>;
   
+  // Generates and stores an identity key, indexed via the following key
+  // account-domain-isLimited
+  // Generates and returns SIWE message for said identity key
+  public abstract getSiweMessage(params: {
```
chris13524: Maybe `createSiweMessage` or `prepareSiweMessage` would be a better name?

Also would there be a desire to extract the identity key creation/storage into a separate function in-case they want to generate the SIWE statement themselves? @Elyniss you had a convo with Trust about this. IMO doesn't seem useful since the SIWE message is pretty fixed anyway.

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -56,17 +58,32 @@ abstract class Client {
...
+    account: string;
+    domain: string,
+    isLimited?: boolean,
```
chris13524: Can we take this opportunity to change this to `allApps?: boolean = false` rather than `isLimited?: boolean = true`. It's more confusing as-is and the statements and links all talk about this app or all apps, not limited vs unlimited. Docs would be a lot easier to understand if this was changed. I had added this to my feature requests here: https://github.com/WalletConnect/web3inbox-widget/issues/40#issuecomment-1821717867

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -56,17 +58,32 @@ abstract class Client {
...
+  // - Verifies signature by reconstructing formatted message 
+  // - Once verified, registers the identity key on the keyserver
   // - Calls `wc_notifyWatchSubscriptions`, watches for future subscription updates, and subscribes to all notification topics
   // Returns the Identity Key public key. Method should throw 'signatureRejected' if any errors comes from onSign promise. 
   public abstract register(params: {
     account: string;
-    private?: boolean;
+    signature: string;
     domain: string,
     isLimited?: boolean,
```
chris13524: Passing around the same `account`, `domain`, and `isLimited` to multiple function calls feels error-prone. Not strongly opposed to it, but I wonder if we should simplify this somehow e.g. by using the identity key's public key or something?

devceline: I debated this too, but it's guarded by the fact that they are used to regenerate the SIWE message and verifying against it as specified in the comment

docs/specs/clients/notify/client-sdk-api.md
devceline: Using `formatSiweMessage` to match what we have with Auth. 

I think generating the identity key in a separate function is overkill and will require more orchestration and ruin devex for not much benefit.

docs/specs/clients/notify/client-sdk-api.md
Elyniss: Yeah, we should keep it, imo

docs/specs/clients/notify/client-sdk-api.md
chris13524: `format` doesn't feel like enough "action" to me. It generates an identity key and saves it which is somewhat important. It's like an impure preparation step, not just a pure formatting function.

devceline: What do you think about `prepareRegistration`? 

Anything ending `SiweMessage` is still deceiving by that logic.

Elyniss: > Also would there be a desire to extract the identity key creation/storage into a separate function in-case they want to generate the SIWE statement themselves? @Elyniss you had a convo with Trust about this. IMO doesn't seem useful since the SIWE message is pretty fixed anyway.

it's not fixed as statement is changed.

I personally prefer `formatSiweMessage`

Elyniss: On second thought, the identity key registration is what makes me want to lean towards `prepareRegistration`. I'm fine with the latest changes

Elyniss (approved):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -56,17 +58,32 @@ abstract class Client {
     encryptedMessage: string,
   }): Promise<NotifyMessage>;
   
+  // Generates and stores an identity key, indexed via the following key
```
flypaper0: I'm not sure that we're need to store identity key before it was registered

Elyniss: We will have to add a flag into local storage that identity is not register on server or something. But we need to generate the key before we generate the message. And we need the message to register the identity key in keys server

flypaper0: Agree that we should generate it here, but should we store it ..

chris13524: If we consider returning the identity key instead of storing it, could we store it in-memory rather than localstorage instead?

Elyniss: @llbartekll we decided that the `prepareRegistration` returns `cacao.payload`, `identityPrivateKey` and `siwe message` to sign. Later on `register` we need that `cacao.payload` to verify the signature and `identityPrivateKey` to recreate `identityPublicKey` and register it in Keys Server

Elyniss (approved):
LGTM

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -56,17 +58,33 @@ abstract class Client {
...
+    message: string
+  };
+  
   // "Logs in" this client to notifications for the specified account. This involves:
-  // - Generating, signing, saving to local stroage, and registering an Identity Key, if one didn't already exist in local storage
+  // - Verifies signature by reconstructing formatted message using provided payload
+  // - Expected payload comes from prepareRegistration
+  // - Once verified, registers the identity key on the keyserver
   // - Calls `wc_notifyWatchSubscriptions`, watches for future subscription updates, and subscribes to all notification topics
   // Returns the Identity Key public key. Method should throw 'signatureRejected' if any errors comes from onSign promise.
```
chris13524:
```suggestion
  // Returns the Identity Key public key. Method should throw 'signatureRejected' if any errors comes from onSign promise. 
```

Should this and the response value be removed now?

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -56,17 +58,33 @@ abstract class Client {
...
-    account: string;
-    private?: boolean;
-    domain: string,
-    isLimited?: boolean,
-    onSign: (message: string) => Promise<Cacao.Signature>
+    registerParams: NotifyRegistrationParams,
+    signature: string;
   }): Promise<string>;
+  
+  // Returns true if account has a valid and up to date registration, false otherwise
```
chris13524, line 84: Specify what level of functionality isRegistered=false has. I think read-only actions should work to improve UX, but engagements should result in needing to re-register. A simple implementation would check `isRegistered` and if false will require user to re-register to continue using the app, but a more advanced implementation could only request the signature when the user requests some action.

```suggestion
  // Returns true if account has a valid and up to date registration, false otherwise
  // If false, the client will still receive notifications from this account, but any actions (e.g. updating a subscription or creating a new one) will be rejected internally by the SDK
```

Perhaps we also need two boolean responses to accurately convey the state, how would a dapp user know if their SDK is still functioning, but the statement is just stale?

docs/specs/clients/notify/client-sdk-api.md
devceline: `signatureRejected` definitely, but we can still return identity key as a form of success

docs/specs/clients/notify/client-sdk-api.md
devceline: I think making that distinction is over-engineering. If `isRegistered` is false they need to/should register. Adding an extra boolean state or making `isRegistered` return a struct will be a bit confusing

docs/specs/clients/notify/client-sdk-api.md
chris13524: > signatureRejected definitely

No because there is no onSign promise anymore

> we can still return identity key as a form of success

Seems useless, but ok

docs/specs/clients/notify/client-sdk-api.md
chris13524: Not over-engineering; if anything over-UX friendly. Otherwise they are effectively logged out if the statement changes.

If we don't want this state, then we should specify that `isRegistered` returning false means that the watch subscriptions aren't being watched, and then you'd also stop receiving notifications until you re-sign?

docs/specs/clients/notify/client-sdk-api.md
devceline: yes, sorry I was unclear `definitely` as in definitely should be removed, already committed removing it

docs/specs/clients/notify/client-sdk-api.md
devceline: As someone who will have to implement this into web3inbox, I have too many choices at this point. Let's assume we split the states. 

1. `isRegistered`: has a statement/identity key. Does not matter if it is stale
2. `hasStaleStatement`: has a stale statement

When do I show the signature modal on web3inbox? 
1. `!isRegistered`
2. `isRegistered && hasStaleStatement`

Because that it is effectively what people creating UIs will have to think about. What would we be recommending?

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -35,38 +37,54 @@ abstract class Client {
...
-  // delete notify message
-  public abstract deleteNotifyMessage(params: {
+  // delete notification
+  public abstract deleteNotification(params: {
     id: number,
   }): Promise<void>;
   
   // decrypt notify subscription message
-  public abstract decryptMessage(params: {
+  public abstract decryptNotification(params: {
```
llbartekll, line 56: it should not be a client method
could we do it like we do in w3w there the method is a part of an interface and depending on implementation client can optionally conform to this interface?

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -35,38 +37,54 @@ abstract class Client {
...
+  // Returns the Identity Key public key. 
   public abstract register(params: {
-    account: string;
-    private?: boolean;
-    domain: string,
-    isLimited?: boolean,
-    onSign: (message: string) => Promise<Cacao.Signature>
+    registerParams: NotifyRegistrationParams,
+    signature: string;
   }): Promise<string>;
```
llbartekll, line 82: what's the motivation for the new approach?

devceline: 2 partners have expressed concerns about passing `onSign`. Also, it'll simplify adoption for wallets and frontends if they control signatures. The current implementation on w3i is very complex and involves 2 different events being fired from within `onSign`. 

This allows for a safer and cleaner implementation

docs/specs/clients/notify/client-sdk-api.md
devceline: Can you link to how we do it in w3w? thoughts @Elyniss @chris13524 @flypaper0  ?

llbartekll: sorry I thought i shared in the first comment
https://github.com/WalletConnect/walletconnect-specs/pull/147/files#
just a detail but this method is not part on the client on some platforms(iOS) as client cannot be accessed during notification decryption.

docs/specs/clients/notify/client-sdk-api.md
chris13524: Certainly more complicated to make it better UX.

Let's leave as "If isRegistered is false they need to/should register." for now. If we update the statement again then we can consider if we then need to handle the UX better.

chris13524 (approved):

llbartekll (approved):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -35,38 +37,54 @@ abstract class Client {
...
+  // Generates and returns SIWE message along with
+  // Cacao Payload with identity key
+  public abstract prepareRegistration(params: {
+    account: string;
+    domain: string,
+    // Defaulted to false if not passed
+    allApps?: boolean,
+  }): {
+    registerParams: NotifyRegistrationParams,
+    message: string
```
flypaper0, line 70: Is it better to move `message` into NotifyRegistrationParams?

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -35,38 +37,54 @@ abstract class Client {
...
   // - Calls `wc_notifyWatchSubscriptions`, watches for future subscription updates, and subscribes to all notification topics
-  // Returns the Identity Key public key. Method should throw 'signatureRejected' if any errors comes from onSign promise. 
+  // Returns the Identity Key public key. 
   public abstract register(params: {
-    account: string;
-    private?: boolean;
-    domain: string,
-    isLimited?: boolean,
-    onSign: (message: string) => Promise<Cacao.Signature>
+    registerParams: NotifyRegistrationParams,
```
flypaper0, line 80: We're planning to take account from Cacao.Payload? May be we can pass it as method param?

devceline: Yes in my implementation I'm taking it from payload

PR #178, opened 2023-12-04
fix: make prepareRegistration async

devceline:


Elyniss (approved):

chris13524 (approved):
Can't we annotate the function as `async` too?

devceline:
> Can't we annotate the function as `async` too?

I don't think we do that in the docs since they're very typescript based - we don't care if the function itself is `async` as long as it returns a promise

chris13524:
> we don't care if the function itself is async as long as it returns a promise

Ah true that is how JavaScript works 🤣

PR #179, opened 2023-12-06
feat: hexless account profiles lookup RPC specs

geekbrother:
This PR adds SPECs for the hexless account profiles JSON-RPC for the forward and reverse lookup in the `blockchain-api` service.

Register and update SPECS https://github.com/WalletConnect/walletconnect-specs/pull/184


docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,37 @@
+# Blockchain-API Server
```
chris13524:
```suggestion
# Blockchain API Server
```

geekbrother: It's still a draft, but thx! fixed.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,91 @@
+# Blockchain API Server
+
+## Blockchain RPC
+
+Blockchain RPC API for interacting with different blockchains can be found in a separate `Blockchain RPC` subsection.
```
geekbrother, line 5: We have a [separate section for the Blockchain RPCs](https://github.com/WalletConnect/walletconnect-specs/tree/main/docs/blockchain-rpc)

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,91 @@
...
+```jsonc
+{
+    "name": string,
+    "text": (optional)string,
+    "avatar": (optional)string,
+    "registered": datetime,
+    "updated": datetime,
+    "addresses": {
+        "{chainID}": (H160)string
+    },
```
arein: nice!

arein (approved):
LGTM!

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,91 @@
...
+
+Used to lookup for suggestions for the new name by returning 3 unique usernames.
+
+`GET /v1/profile/suggestion`
+
+#### Success response body:
+
+```jsonc
+{
+    "suggestions": [{name}],
```
chris13524:
```suggestion
    "suggestions": string[],
```

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,91 @@
...
+
+* `name` - registered account profile name.
+* `text` - account profile self description.
+* `avatar` - URL of the avatar for the account.
+* `chainID` - chain id where the lookup address is bind.
+
+#### Response error codes:
+
+* `404 Not Found` - The requested address is not registered for any name.
+* `406 Not Acceptable` - Wrong requested address format.
```
chris13524: 406 is not an appropriate response here, we should either use 400 or 422

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,91 @@
...
+* `404 Not Found` - The requested name is not registered.
+* `406 Not Acceptable` - Wrong requested name format.
+
+### Reverse name lookup
+
+Used to lookup registered account name for the address.
+
+`GET /v1/profile/reverse/{address}`
+
+* `address` - is the address in `H160` format. eg. `0x71C7656EC7ab88b098defB751B7401B5f6d8976F`
```
chris13524: H160 is actually a binary format, address just needs to be a string.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Fixed.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Agree! Changed to the 400 Bad Request.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Removed the `H160` if it's looks confusing.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+
+```jsonc
+{
+    "name": string,
+    "address": string,
+    "text": (optional)string,
+    "avatar": (optional)string,
+    "registered": datetime,
+    "updated": datetime,
+    "chainIDs" []integer
```
geekbrother, line 31: I've changed the response format: 
The address is the same for the Ethereum chainIDs, we just need to register the list that the user allows to work with, so we don't need to duplicate it in the format like:
```
"60": "0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
"1":"0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
```
It's better to expose the list of allowed chainIDs.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+
+```jsonc
+{
+    "name": string,
+    "address": string,
+    "text": (optional)string,
+    "avatar": (optional)string,
+    "registered": datetime,
+    "updated": datetime,
+    "chainIDs" integer[]
```
geekbrother: I've changed the response format: 
The address is the same for the Ethereum chainIDs, we just need to register the list that the user allows to work with, so we don't need to duplicate it in the format like:
```
"60": "0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
"1":"0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
```
It's better to expose the list of allowed chainIDs.

sidebars.js
```diff
@@ -138,6 +138,16 @@ module.exports = {
                 },
               ],
             },
+            {
+              type: "category",
+              label: "Blockchain Server",
```
chris13524:
```suggestion
              label: "Blockchain API",
```

docs/specs/servers/blockchain/readme.md
```diff
@@ -0,0 +1,5 @@
+# Blockchain API Server Overview
```
chris13524, line 1:
```suggestion
# Blockchain API Overview
```

docs/readme.mdx
```diff
@@ -69,6 +69,11 @@ url: `/2.0/specs/servers/archive`
 name: "Verify Server",
 description: "Verify server is used to authenticate the origin of an attestation id which is related to a JSON-RPC payload sent.",
 url: `/2.0/specs/servers/verify`
+},
+{
+name: "Blockchain API Server",
```
chris13524:
```suggestion
name: "Blockchain API",
```

docs/readme.mdx
```diff
@@ -69,6 +69,11 @@ url: `/2.0/specs/servers/archive`
...
+description: "Blockchain API server is used for interacting with different blockchains, resolving identity, ENS interactions and transactions history.",
```
chris13524:
```suggestion
description: "Blockchain API is used for interacting with different blockchains, resolving identity, ENS interactions and transactions history.",
```

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Blockchain RPC
+
+Blockchain RPC API for interacting with different blockchains can be found in a separate `Blockchain RPC` subsection.
+
+## Hexless account profiles API
+
+Email wallet users can register a profile name e.g. `coolname.connect.id` and use it instead of the `0x` address.
+Server API exposes forward lookup for the `name.connect.id` -> `0x` address with chain IDs and reverse lookup for the `0x` -> `name.connect.id`.
+Users can also update the name data including avatar, description text and bind addresses.
```
chris13524: Need to have blank line in-between if you intend for these to be separate paragraphs.

```suggestion
Email wallet users can register a profile name e.g. `coolname.connect.id` and use it instead of the `0x` address.

Server API exposes forward lookup for the `name.connect.id` -> `0x` address with chain IDs and reverse lookup for the `0x` -> `name.connect.id`.

Users can also update the name data including avatar, description text and bind addresses.
```

docs/specs/servers/blockchain/blockchain-server-api.md
chris13524: Does optional mean the field might not exist or it's nullable? Let's make it explicit. Also there is no datetime or integer in JSON, just numbers or strings, so again let's specify that.

Also `chain_ids` as a stylistic preference.

```suggestion
```typescript
{
    "name": string,
    "address": string,
    "text": string | null,
    "avatar": string | null,
    "registered": number,
    "updated": number,
    "chain_ids" number[]
```

chris13524: When we support non-EVM or smart-contract accounts, it would make sense to have multiple address formats supported and different addresses on each chain. Maybe we should do an array of addresses in CAIP-10 format instead of separate address and chain IDs fields?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+    "registered": datetime,
+    "updated": datetime,
+    "chainIDs" integer[]
+}
+```
+
+* `name` - registered account name.
+* `address` - address that bind to the account name.
+* `text` - account profile self description.
+* `avatar` - URL of the avatar for the account.
```
chris13524: Where are we hosting avatars? Is it better to do base64 encoded data here? Then client can display as a data URI.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+    "avatar": (optional)string,
+    "registered": datetime,
+    "updated": datetime,
+    "chainIDs" integer[]
+}
+```
+
+* `name` - registered account name.
+* `address` - address that bind to the account name.
+* `text` - account profile self description.
```
chris13524: What's `text` used for?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+Server API exposes forward lookup for the `name.connect.id` -> `0x` address with chain IDs and reverse lookup for the `0x` -> `name.connect.id`.
+Users can also update the name data including avatar, description text and bind addresses.
+
+### Name lookup
+
+Used to lookup address for the name.
+
+`GET /v1/profile/account/{name}`
+
+* `name` - is the registered name. eg. `coolname.connect.id`
```
chris13524, line 19: What is valid format for this so we can sanitize? Alpha numeric and periods? Max length?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+```jsonc
+{
+    "payload":
+    {
+        "name": string,
+        "text": (optional)string,
+        "avatar": (optional)string,
+        "address": string,
+        "chainIDs": integer[]
+    },
```
chris13524: `payload` should be a serialized string, not an object, to ensure there are no problems reconstructing a consistent signature payload with different JSON libraries.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+* `201 Created` - account name is registered.
+
+#### Response error codes:
+
+* `400 Bad Request` - some parameters in request body were missed.
+* `401 Unauthorized` - signature verifying error.
+
+### Name suggestion
+
+Used to lookup for suggestions for the new name by returning 3 unique usernames.
```
chris13524, line 100: Why 3? Why not 1?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+{
+    "payload":
+    {
+        "name": string,
+        "text": (optional)string,
+        "avatar": (optional)string,
+        "address": string,
+        "chainIDs": integer[]
+    },
+    "signature": string
```
chris13524: Have we considered using a standard SIWE with the payload in the `statement` field instead of a custom signature pattern here? It would allow non-auto sign scenarios (e.g. in the user's mobile wallet), but would be harder to implement. I suppose we could always implement this alternative later.

chris13524 (changes requested):

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,136 @@
...
+* `name` - is the registered name. eg. `coolname.connect.id`
+
+#### Success response body:
+
+```jsonc
+{
+    "name": string,
+    "address": string,
+    "text": (optional)string,
+    "avatar": (optional)string,
```
chris13524: Max length? What are valid chars?

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Hosting images in the Postgres database even as small as 256x256 is a bad idea for the scale and performance that we need. The database size in this case for the 100Mi of users will be monstrous, this is costly, and database pressure (even if using the BLOB type, which is stored out of the table's main index). Zerion, for example, stores images/icons in the S3 storage and just uses links in the API responses.
We definitely should store images and other blobs outside of the DB (s3, storj, ipfs etc.).

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Should be changed to `bio`, as a more self-describing name.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: We agreed on this in our [kickoff meet](https://www.notion.so/walletconnect/ENS-Tech-Design-Kickoff-73f06bfc08c04da8ad1407f17063d8c7?pvs=4#280429d86a3248ebb4af831e364dbbba).

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: I think we should postpone other blockchains for the `v2` and stick to the Eth for the MVP and `v1`, because the scope for that is much bigger (different formats, validation, signing, etc.) I'll make the database schema with keeping this in mind so we can add another chains without much effort in it later, but I think we should postpone with this in the server implementation for now.

sidebars.js
geekbrother: Done. thx.

docs/specs/servers/blockchain/readme.md
geekbrother: Done, thx.

docs/readme.mdx
geekbrother: Done, thx.

docs/readme.mdx
geekbrother: Done, thx.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Makes sense to make it more readable. Done, thanks!

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Changed to `bio`.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Makes sense! Added constraints description.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Changed.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Makes sense! Added constraints description.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,143 @@
...
+}
+```
+
+* `payload` - serialized string that should contain the following fields:
+
+    * `name` - is the name to register or update.
+        * Alphabetic characters (A-Z a-z), numeric characters (0-9), the minus sign (-) and the period (.)
+        * Max 255 characters length.
+    * `bio` - account profile self description.
+        * Max 140 characters length.
```
geekbrother: Bio max length size inspired by Telegram (140) and Discord (190).

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Makes sense to change it to this format. Done.

docs/specs/servers/blockchain/blockchain-server-api.md
chris13524: S3 probably. So probably need the API to accept file uploads and then it will push to S3? Also handling the deletion of the S3 object case if a new avatar is uploaded. What about consistency if upload works but save to Postgres fails?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Success response body:
+
+```typescript
+{
+    "address": string,
+    "name": string,
+    "bio": string | null,
+    "avatar": string | null,
+    "chain_ids": integer[]
```
chris13524:
```suggestion
    "chain_ids": number[]
```

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,143 @@
...
+{
+    "payload": string,
+    "signature": string
+}
+```
+
+* `payload` - serialized string that should contain the following fields:
+
+    * `name` - is the name to register or update.
+        * Alphabetic characters (A-Z a-z), numeric characters (0-9), the minus sign (-) and the period (.)
```
chris13524: Are `CoolDerek` and `coolderek` same or different names? Could this be a security problem or something if they type lowercase expecting it to be the same? DNS considers different caps the same, and I assume ENS does too, if not only allowing lowercase.

chris13524 (changes requested):

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Makes sense, yes DNS is a case agnostic so we should be the same here.

docs/specs/servers/blockchain/blockchain-server-api.md
chris13524: I think it would be easy enough to use CAIP-10 now and should be future-proof, for example the `eip155:1:0xFF` is the account ID. This is how Notify Server does it, and it only supports EVM currently. Assert eip155 prefix to be EVM during signature validation, and then extract the chain ID from that. Can return as an array of CAIP-10 addresses.

geekbrother: Ok, makes sense to change this in a resolver now and think more about the register/update specs a bit later.
I've made changes to use CAIP-10 and multichain, and extracted the register/update to a separate PR https://github.com/WalletConnect/walletconnect-specs/pull/184 to unblock the implementation of the resolving for now.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Good question to think about. I've extracted the register/update to https://github.com/WalletConnect/walletconnect-specs/pull/184 to unblock the resolving implementation and we can move to the https://github.com/WalletConnect/walletconnect-specs/pull/184 for this.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: I should make changes on this for the multichain sign as well. 
I've extracted register/update into a separate https://github.com/WalletConnect/walletconnect-specs/pull/184 to discuss further and unblock the resolver itself with allowlisted data for now.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Fixed. Thanks 🙏

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,108 @@
...
+#### Success response body:
+
+```typescript
+{
+    "name": string,
+    "bio": string | null,
+    "avatar": string | null,
+    "registered": number,
+    "updated": number,
+    "addresses": object[]
```
chris13524:
```suggestion
    "addresses": { address: string, updated: number }[]
```

chris13524 (approved):

geekbrother:
In [c353ac](https://github.com/WalletConnect/walletconnect-specs/pull/179/commits/c353ac3b26ae6221f20002259a4b6b14898897d7) moved the `avatar` and `bio` to the `attributes` object, as we should extend it in the future according to our design.

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -0,0 +1,112 @@
...
+* `registered` - unix time stamp of the name registration.
+* `updated` - unix time stamp  of the last name data update.
+* `addresses` - list of objects with name addresses data:
+    * `blockchain` - blockchain of the address. e.g. `eip155` for the Ethereum.
+    * `chain_id` - chain ID of the address. Can be NULL for the mainnet.
+    * `address` - address according to the blockchain format.
+    * `created` - unix time stamp when the address was added.
+* `attributes` - key value object of the name attributes:
+    * `avatar` - (Optional) avatar url.
+    * `bio` - (Optional) account profile self description.
```
chris13524, line 45: Do these need to be moved?

PR #181, opened 2023-12-07
Make `isRegistered` take same params as `prepareRegistration` to check registration status accurately

devceline:
Make `isRegistered` take same params as `prepareRegistration` to check registration status accurately

### DIFF
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -83,7 +83,9 @@ abstract class Client {
   
   // Returns true if account has a valid and up to date registration, false otherwise
   public abstract isRegistered(params: {
-    account: string
+    account: string,
+    allApps: boolean,
+    domain: string
   }): boolean
 
   // "Logs out" this client from any notifications for the specified account. This involves:


Elyniss (approved):

flypaper0 (approved):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -83,7 +83,9 @@ abstract class Client {
   
   // Returns true if account has a valid and up to date registration, false otherwise
   public abstract isRegistered(params: {
-    account: string
+    account: string,
+    allApps: boolean,
```
chris13524: Should this have the same default as above?

```suggestion
    // Defaulted to false if not passed
    allApps?: boolean,
```

devceline: I don't think it should be optional personally - you're checking for permissions essentially. 

Is `account` registered under `domain` to access a single app or `allApps`

chris13524: I don't see your argument, I don't see anything wrong with having a default in what you're checking for. I don't want 90% of dapp developers to have to set this to `false` when they don't for the other one.

Elyniss: I think that Kotlin and Swift clients it's better to set this to true instead of false

chris13524: Default true on Swift and Kotlin makes sense to me since those are pretty much exclusive to wallets, but dapps are still possible. But for JS it's more tricky because there are React Native wallets.

I would almost rather not have a default, than to default to true.

Also once we pick a default, we cannot change it even through a breaking change because you risk people not reading the migration doc and just upgrading major versions.

devceline:
@Elyniss @chris13524 @flypaper0 added a default to false

chris13524 (approved):

PR #184, opened 2023-12-08
feat: hexless account profiles register and update RPC specs

geekbrother:
Registering hexless account name and updating it.
Adding and updating addresses for the name.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-server-api.md
+++ b/docs/specs/servers/blockchain/blockchain-server-api.md
@@ -10,6 +10,8 @@ Email wallet users can register a profile name e.g. `coolname.connect.id` and us
 
 Server API exposes forward lookup for the `name.connect.id` -> `0x` address with chain IDs and reverse lookup for the `0x` -> `name.connect.id`.
 
+Users can also update the name data including avatar, bio and bind addresses.
+
 ### Name lookup
 
 Used to lookup address for the name.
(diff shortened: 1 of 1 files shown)


chris13524 (approved):
LGTM

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -106,3 +108,45 @@ Used to lookup for suggestions for the new name by returning 3 unique usernames.
...
+```
+
+* `message` - JSON serialized string that should contain the following fields:
+
+    * `name` - is the name to register or update.
+        * Alphabetic characters (A-Z a-z), numeric characters (0-9), the minus sign (-) and the period (.)
+        * Max 255 characters length.
+    * `address` - address that bind to the account name.
+        * Max 100 characters length.
+    * `timestamp` - current unixtime timestamp. The signature is valid for 30 seconds.
```
geekbrother: Adding the 30 seconds validity of the signature, to limit replay attack.

chris13524: Might be a bit short to accommodate clock drift, I'd do 5 minutes to be safe.

geekbrother: The reference for the 30 seconds for my decision is the [TOTP](https://ru.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm) it's time-based and it has a 30 seconds gap for the sync.

PR #187, opened 2023-12-23
fix: add sequence number of subscriptions revision

chris13524:
[Context](https://walletconnect.slack.com/archives/C044SKFKELR/p1703322752337749)

## Problem 

There is a race condition if multiple subscription updates happen at the same time that a client will receive them out-of-order and display an outdated version.

## Solution

Notify Server will now store a version for the subscriptions list. I.e. everytime a subscription is updated (in a way that would trigger subscriptions changed e.g. a subscribe, update, or delete), then this sequence number will also be incremented.

The sequence number is sent to the account alongside the list of subscriptions which clients will then store with their local list of subscriptions. Whenever the client processes a `sbs` value it is to first check the sequence number, and if it is smaller than the current number then it is to disregard the update.


flypaper0 (approved):
lgtm

devceline (approved):

PR #188, opened 2024-01-03
Add image urls scopes

devceline:
- Add imageUrls to scopes as we want to use images from notification types, nor from the `icon` property of notifications
- Add `id` and `name` since they are already implemented

### DIFF
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -30,7 +30,7 @@ abstract class Client {
   // query notification types available for a dapp domain
   public abstract getNotificationTypes(params: {
     appDomain: string,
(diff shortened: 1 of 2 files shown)


Elyniss (commented):
Overall change lgtm, however, I noticed we also should change it on `getNotificationTypes()` response. Could we call this structure ~`NotifyAvailableTypes`~ `NotifyNotificationTypes`, rename the `getNotificationTypes()` response to `NotifyNotificationTypes` and kill two birds with one stone? 

To not have breaking changes, not change the clients if they have different naming? 

---

Current specs for context:


```
  // query notification types available for a dapp domain
  public abstract getNotificationTypes(params: {
    appDomain: string,
  }): Promise<NotifyAvailableTypes>
```

Elyniss (approved):

PR #200, opened 2024-01-30
chore(web3modal-api): add chains filter param

Cali93:
Follow up PR to:
- https://github.com/WalletConnect/api-web3modal/pull/43#issue-2107190000


bkrem (approved):
Nice :100: good call on updating the specs reference I forgot about this 🙏

PR #201, opened 2024-02-05
chore: update to a desirable client api

devceline:
- remove `getUnreadCount` as it would be redundant since we should supply in read count in get subscriptions call 
- simplify `notificiations_changed_event` so it can be easily used for diffing state cached by the client or user of the sdk
- add `unreadCount` to subscriptions
- Remove `getNotification` as it would be redundant. If someone wanted to display details of a selected notification, they can retrieve it from the existing list of notifications

### DIFF
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -57,7 +57,7 @@ abstract class Client {
   public abstract markNotificationsAsRead(params: {
     topic: string,
     notificationIds: string[],
-	all?: boolean
+    all?: boolean
   }): Promise<NotifyNotificationRecord>
 
   // delete active subscription
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -51,22 +51,14 @@ abstract class Client {
     hasMoreUnread: boolean,
   }>
 
-  // get notification by ID
-  public abstract getNotification(params: {
-    topic: string,
-    id: string,
-  }): Promise<NotifyNotificationRecord>
-
   // mark notification as read
+  // if `all` is marked as true, 
+  // the array of IDs is ignored and all notifications are marked as read
   public abstract markNotificationsAsRead(params: {
     topic: string,
-    ids: string[],
+    notificationIds: string[],
+	all?: boolean
   }): Promise<NotifyNotificationRecord>
-
-  // returns how many notifications are unread
-  public abstract getUnreadNotificationsCount(params: {
-    topic: string,
-  }): Promise<number>
 
   // delete active subscription
   public abstract deleteSubscription(params: {
@@ -131,7 +123,7 @@ abstract class Client {
   public abstract on("notify_notification", (notification: NotifyNotificationRecord, metadata: Metadata) => {}): void;
 
   // Listen for when an existing notification has been changed
-  public abstract on("notify_notification_changed", (notification: NotifyNotificationRecord, metadata: Metadata) => {}): void;
+  public abstract on("notify_notifications_changed", (notifications: Record<string, { read: boolean }>) => {}): void;
 
   // for wallet to listen for result of notify subscription update
   public abstract on("notify_update", (result: NotifySubscription | Error) => {}): void;
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -54,7 +54,7 @@ abstract class Client {
   // mark notification as read
   // returns true if operation was a success. 
   // For now, `read` can only be true. 
-  public abstract updateNotificationReadState(params: {
+  public abstract updateNotificationsReadState(params: {
     topic: string,
     notificationIds: string[],
 	read?: true
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -52,10 +52,13 @@ abstract class Client {
   }>
 
   // mark notification as read
-  public abstract markNotificationsAsRead(params: {
+  // returns true if operation was a success. 
+  // For now, `read` can only be true. 
+  public abstract updateNotificationReadState(params: {
     topic: string,
     notificationIds: string[],
-  }): Promise<NotifyNotificationRecord>
+	read?: true
+  }): Promise<boolean>
 
   // delete active subscription
   public abstract deleteSubscription(params: {
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -52,12 +52,9 @@ abstract class Client {
   }>
 
   // mark notification as read
-  // if `all` is marked as true, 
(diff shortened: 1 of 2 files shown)


docs/specs/clients/notify/data-structures.md
```diff
@@ -65,6 +66,7 @@ NotifyServerSubscription[]
   account: Account, // CAIP-10 account
   scope: string[], // Array of notification types enabled for this subscription
   expiry: number, // Unix timestamp of expiration
+  unreadCount: number
```
Elyniss, line 69: Should we also add all notifications count?

docs/specs/clients/notify/data-structures.md
```diff
@@ -13,6 +13,7 @@
   "metadata": Metadata,
   "scope": Record<string, {description: string, enabled: boolean}>,
   "expiry": number,
+  "unreadNotificationCount": number
 }
 ```
```
Elyniss, line 19: I would also add to client facing models read flag, namely NotifyNotificationRecord

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -131,7 +123,7 @@ abstract class Client {
   public abstract on("notify_notification", (notification: NotifyNotificationRecord, metadata: Metadata) => {}): void;
 
   // Listen for when an existing notification has been changed
-  public abstract on("notify_notification_changed", (notification: NotifyNotificationRecord, metadata: Metadata) => {}): void;
+  public abstract on("notify_notifications_changed", (notifications: Record<string, { read: boolean }>) => {}): void;
```
Elyniss, line 126: Why not return whole NotifyNotificationRecord instead? If we ever add any more data/states to this event we would have to have a migration

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -51,23 +51,15 @@ abstract class Client {
...
-  }): Promise<NotifyNotificationRecord>
-
   // mark notification as read
+  // if `all` is marked as true, 
+  // the array of IDs is ignored and all notifications are marked as read
   public abstract markNotificationsAsRead(params: {
     topic: string,
-    ids: string[],
+    notificationIds: string[],
+    all?: boolean
```
Elyniss: I'm not a fan of this all flag. What if clients don't have all messages and mark this flag?

devceline: that's actually a really good point, it was brought up in the meeting but you're absolutely right we could have a situation where someone marks all notifications as read, even the ones they haven't received yet if their client isn't refetching notifications every time a new one pops up.

I was writing these docs with the hooks in mind, but we can't count on all users of the client doing that

docs/specs/clients/notify/data-structures.md
devceline: What would be the use case of that

docs/specs/clients/notify/data-structures.md
devceline: that exists already

docs/specs/clients/notify/client-sdk-api.md
devceline: It shows that the only thing that can be changed about a notification is its read status. If we add more state later on, we can update the specs no?

docs/specs/clients/notify/data-structures.md
Elyniss: right

docs/specs/clients/notify/data-structures.md
Elyniss: Yeah true, we will add it when necessary

flypaper0 (approved):

docs/specs/clients/notify/data-structures.md
chris13524, line 69: Why is this named different from `unreadNotificationCount`? Should we prefer one or the other?

Elyniss: Dont have strong opinion on this

Elyniss (approved):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -51,23 +51,12 @@ abstract class Client {
...
-    topic: string,
-    id: string,
-  }): Promise<NotifyNotificationRecord>
-
   // mark notification as read
   public abstract markNotificationsAsRead(params: {
     topic: string,
-    ids: string[],
+    notificationIds: string[],
   }): Promise<NotifyNotificationRecord>
```
Elyniss: How about we rename it to changeNotificationsReadState and add boolean flag

devceline: But we won't have the ability to mark as unread yet. Hm. Maybe it can be

```typescript
public abstract changeNotificationReadState(params: {
  topic: string,
  notificationIds: string[]
  read?: true
})
```

Where read is optional and if provided has to be `true`. Down the line it can change to `read: boolean` so it's not a breaking change?

cc @chris13524

Elyniss: Makes sense

chris13524: > Down the line it can change to read: boolean so it's not a breaking change?

Changing it to non-optional is a breaking change.

chris13524: Aside: We could make the RPC method [name] even more generic, e.g. `updateNotification`.

But for this function name I'm fine with the proposal.

devceline: wouldn't have to change it to non optional, it would just be `read = true` by default

Elyniss: yeah, lets remove optional from above suggestion

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -51,22 +51,14 @@ abstract class Client {
...
-
-  // returns how many notifications are unread
-  public abstract getUnreadNotificationsCount(params: {
+  // returns true if operation was a success. 
+  // For now, `read` can only be true. 
+  public abstract updateNotificationReadState(params: {
     topic: string,
-  }): Promise<number>
+    notificationIds: string[],
+	read?: true
```
Elyniss: please fix formatting (tabs)

Elyniss (changes requested):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -51,22 +51,14 @@ abstract class Client {
...
-  public abstract markNotificationsAsRead(params: {
-    topic: string,
-    ids: string[],
-  }): Promise<NotifyNotificationRecord>
-
-  // returns how many notifications are unread
-  public abstract getUnreadNotificationsCount(params: {
+  // returns true if operation was a success. 
+  // For now, `read` can only be true. 
+  public abstract updateNotificationReadState(params: {
```
Elyniss:
```suggestion
  public abstract updateNotificationsReadState(params: {
```

Elyniss (changes requested):

Elyniss (approved):

PR #204, opened 2024-02-08
feat(ens): name address update

geekbrother:
This PR adds the specification for updating the hexless account name address to give users the ability to update addresses after the name registration.

To authorize the name ownership we will use the same technique as in the name registration signed message with payload.
The recovered address will be compared to the address from the name entity to check the legitimate owner update.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-server-api.md
+++ b/docs/specs/servers/blockchain/blockchain-server-api.md
@@ -193,6 +193,47 @@ The POST request body should be in JSON format with the folowing structure:
 * `400 Bad Request` - some parameters in request body were missed or wrong.
 * `401 Unauthorized` - signature verification or name address ownership error.
 
+### Updating name addresses
+
+Used to update or add name addresses.
+
+`POST /v1/profile/account/{name}/address`
+
+* `name` - is the registered name. eg. `coolname.wc.ink`
+
+#### Request body:
+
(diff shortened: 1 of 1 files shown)


docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -193,6 +193,47 @@ The POST request body should be in JSON format with the folowing structure:
...
+#### Request body:
+
+The POST request body should be in JSON format with the folowing structure:
+
+```typescript
+{
+    "message": string,
+    "coin_type": string,
+    "signature": string,
+    "address":string,
```
chris13524, line 213:
```suggestion
    message: string,
    coin_type: string,
    signature: string,
    address: string,
```

chris13524 (approved):
LGTM

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: I see that we have such a convention in [notify as well](https://github.com/WalletConnect/walletconnect-specs/blob/65b0deaf1738bfd53e8be0de98acbb7928d92560/docs/specs/servers/notify/notify-server-api.md?plain=1#L17-L22) 👀 so if we are going to change it, it's better to change everywhere in a followup.

PR #206, opened 2024-02-23
feat(blockchain-api): account balances

geekbrother:
This PR adds specifications for the Account Balances request and response to the Blockchain-API service.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-server-api.md
+++ b/docs/specs/servers/blockchain/blockchain-server-api.md
@@ -296,12 +296,12 @@ Used to lookup address asset balances
 
 #### Success response body:
 
-* `positions` - list of address positions with the following attributes:
+* `balances` - list of address balances with the following attributes:
     * `name` - asset name. e.g. `Ethereum`
     * `symbol` - native asset or ERC-20 symbol. e.g. `ETH`
     * `chainId` - CAIP-10 chain id. e.g. `eip155:1`
     * `value` - asset value in currency calculated from the request argument.
-    * `price` - asset price in currency calculated from the request argument.
+    * `price` - price of a single unit of the asset in currency calculated from the request argument.
     * `quantity` - asset quantity object:
         * `decimals` - decimal number precision of the quantity.
         * `numeric` - numeric representation of the asset quantity.
--- a/docs/specs/servers/blockchain/blockchain-server-api.md
+++ b/docs/specs/servers/blockchain/blockchain-server-api.md
@@ -276,3 +276,37 @@ The POST request body should be in JSON format with the following structure:
 
 * `400 Bad Request` - some parameters in request body were missed or wrong.
 * `401 Unauthorized` - projectID verification error.
+
+## Account 
+
+### Balance
+
+Used to lookup address asset balances
+
+`GET /v1/account/{address}/balance`
+
+#### Path parameters
+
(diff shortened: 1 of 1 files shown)


docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -276,3 +276,37 @@ The POST request body should be in JSON format with the following structure:
...
+* `chain` - (Optional) filter assets by CAIP-10 chain id.
+
+#### Success response body:
+
+* `positions` - list of address positions with the following attributes:
+    * `name` - asset name. e.g. `Ethereum`
+    * `symbol` - native asset or ERC-20 symbol. e.g. `ETH`
+    * `chainId` - CAIP-10 chain id. e.g. `eip155:1`
+    * `value` - asset value in currency calculated from the request argument.
+    * `price` - asset price in currency calculated from the request argument.
```
arein:
```suggestion
    * `price` - price of a single unit of the asset in currency calculated from the request argument.
```

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -276,3 +276,37 @@ The POST request body should be in JSON format with the following structure:
 
 * `400 Bad Request` - some parameters in request body were missed or wrong.
 * `401 Unauthorized` - projectID verification error.
+
+## Account 
+
+### Balance
+
+Used to lookup address asset balances
```
arein: can you cover whether this includes non-fungible assets?

arein (approved):

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -276,3 +276,37 @@ The POST request body should be in JSON format with the following structure:
...
+* `address` - is the address for lookup. eg. `0xab16a96D359eC26a11e2C2b3d8f8B8942d5Bfcdb`
+
+#### Query parameters
+
+* `currency` - currency to calculate the assets value. e.g. `USD`
+* `chain` - (Optional) filter assets by CAIP-10 chain id.
+
+#### Success response body:
+
+* `positions` - list of address positions with the following attributes:
```
chris13524: Assuming we are just forwarding to another API, but this API would return _all_ positions which could be unnecessary and large, and would also not support a custom token ID/address

chris13524 (approved):
Some comments on the PR that would be great to clarify in the spec, but otherwise looks like good approach so approving

docs/specs/servers/blockchain/blockchain-server-api.md
bkrem: Seems a bit disjointed from the naming of the endpoint, wouldn't `balances` fit better? My first instinct reading `positions` is to think trading.

bkrem (approved):
🚀

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Makes sense to me to change it to `balances`. It's changed. Thanks for the suggestion!

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Resolved in https://github.com/WalletConnect/walletconnect-specs/pull/206/commits/393a30b86b9852bea474372754bcae4482fa1e4b

PR #213, opened 2024-03-21
fix(balance): returning list of token implementation addresses

geekbrother:
This PR makes changes to the blockchain-api balance endpoint to return a list of token implementations instead of just one, because the token can have multiple implementations on multiple chains.


nopestack (approved):

geekbrother:
Closing this API change as we decided to go forward with the same approach with the single implementation.

PR #216, opened 2024-04-03
feat(blockchain-api): fungible price endpoint

geekbrother:
This is a proposal for the fungible price endpoint SPEC for the Blockchain-API.

The new endpoint `/v1/fungibles/price` should respond with the price of the fungible asset calculated in the requested currency. 
The fungible assets are filtered by the CAIP-10 implementation address.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-server-api.md
+++ b/docs/specs/servers/blockchain/blockchain-server-api.md
@@ -359,6 +359,43 @@ Used to lookup fungible assets balances
 * `202 Accepted` - The data for the address is not fulfilled yet, re-trying is needed in up to 180 seconds.
 * `400 Bad request` - Wrong requested arguments format.
 
+## Fungibles
+
+### Price
+
+Lookup for the current price of the fungible assset item:
+
+`POST /v1/fungible/price`
+
+#### Request body:
+
+The POST request body should be in JSON format with the folowing structure:
+
+* `projectId` - The project identifier.
+* `currency` - Currency for the price calculation e.g. `usd`.
+    * Supported currencies: `usd`, `eur`, `gbp`, `aud`, `cad`, `inr`, `jpy`, `btc`, `eth`.
+* `addresses` - Asset's contract addresses array of strings in [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md) format.
+
+#### Success response body:
+
+```typescript
+{
+    "fungibles": object[],
+}
+```
+
+* `fungibles` - List of objects with fungibles on the requested address:
+    * `address` - Asset implementation address.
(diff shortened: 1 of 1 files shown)


docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -331,3 +331,36 @@ Used to lookup fungible assets balances
...
+
+#### Success response body:
+
+```typescript
+{
+    "fungibles": object[],
+}
+```
+
+* `fungibles` - List of objects with fungibles on the requested address:
```
tomiir, line 387: Why a list of objects as response? Wouldnt it be a single entry?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -331,3 +331,36 @@ Used to lookup fungible assets balances
...
+Lookup for the current price of the fungible assset item:
+
+`GET /v1/fungibles/price`
+
+#### Path parameters
+
+* `projectId` - The project identifier.
+* `currency` - Currency for the price calculation e.g. `usd`.
+    * Supported currencies: `usd`, `eur`, `gbp`, `aud`, `cad`, `inr`, `jpy`, `btc`, `eth`.
+* `address` - Asset's contract address in [CAIP10](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md) format.
```
tomiir: @enesozturk is the address useful for you? 
Wouldn't it be simpler to send a symbol?

Requiring an address here would require the ui to somehow have access to it which will probably be an extra integration on FE. 

Wondering if chainid + symbol would be better

enesozturk: @tomiir I think the address is required by the API we are using. We send address of the tokens to fetch prices. I'm not sure if we are using 1Inch API or Zerion. In 1Inch price API, the prices endpoint gets address array instead of one address and returns the array as response.

We need this features in swaps to fetch our owned tokens balances at once. @geekbrother

geekbrother: > @tomiir I think the address is required by the API we are using.

That's correct, we need it for the API and I think one symbol can have different implementations (different chains/contracts) ?

> In 1Inch price API, the prices endpoint gets address array instead of one address and returns the array as response.

Let's change this to the array of addresses to use this in swaps as well.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Is it possible that different fungibles are deployed at the same address on different chains (from Zerion API doc).

enesozturk (approved):

PR #222, opened 2024-04-10
feat(relay): `irn_batchRequest` method

heilhead:
This adds the `irn_batchRequest` method which would allow to combine multiple requests into a single one, specifying execution mode (either concurrent or sequential). See [this slack conversation](https://walletconnect.slack.com/archives/C03SMNKLPU0/p1711981001065359) for context.

### DIFF
--- a/docs/specs/servers/relay/relay-server-rpc.md
+++ b/docs/specs/servers/relay/relay-server-rpc.md
@@ -424,6 +424,32 @@ Used to unregister an active watch webhook corresponding to a webhookId.
 }
 ```
 
+### Batch Request
+
+Allows combining multiple requests into a single one. Used to save on round-trip latency.
+
+```jsonc
+// Request (client->server)
+{
+  "id" : "1",
+  "jsonrpc": "2.0",
+  "method": "irn_batchRequest",
+  "params" : {
+    "batch": Request[],
+    "mode": "sequence" | "concurrent"
+  }
+}
+
+// Response (server->client)
+{
+  "id" : "1",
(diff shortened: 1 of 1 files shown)


docs/specs/servers/relay/relay-server-rpc.md
```diff
@@ -424,6 +424,32 @@ Used to unregister an active watch webhook corresponding to a webhookId.
...
+
+```jsonc
+// Request (client->server)
+{
+  "id" : "1",
+  "jsonrpc": "2.0",
+  "method": "irn_batchRequest",
+  "params" : {
+    "batch": Request[],
+    "mode": "sequence" | "concurrent"
```
xDarksome, line 439: I thought we had a use-case where we need both `sequence` and `concurrent` at the same time. 
Like:
- execute 3 requests concurrently
- afterwards execute another request

Isn't it the case?

heilhead: Correct. So with the suggested implementation you'd combine requests like this:
```js
{
  "method": "irn_batchRequest",
  "params": {
    "mode": "sequence",
    "batch": [
      // First execute a batch of requests concurrently.
      {
        "method": "irn_batchRequest",
        "params": {
          "mode": "concurrent",
          "batch": [
            // Concurrent requests, e.g. subscriptions.
          ]
        },
        ...
      },

      // Then another request once the first batch is processed.
      {
        "method": "irn_publish",
        ...
      },
    ]
  },
  ...
}
```

PR #229, opened 2024-05-07
chore: add prepare registration with recaps spec

devceline:
Recaps support has been in the notify SDKs for a while but we have not officially specced it out yet.

Here's the spec, mimicking the current implementation on JS.

### DIFF
--- a/docs/specs/clients/notify/client-sdk-api.md
+++ b/docs/specs/clients/notify/client-sdk-api.md
@@ -67,6 +67,20 @@ abstract class Client {
     topic: string,
     encryptedMessage: string,
   }): Promise<NotifyNotification>;
+  
+
+  // Generates and returns incomplete CacaoPayload
+  // To be handled in recaps based authentication.
+  // This enables one-click-auth based authentication to
+  // also enable permissions for notify API.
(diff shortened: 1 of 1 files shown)


chris13524 (approved):

docs/specs/clients/notify/client-sdk-api.md
```diff
@@ -67,6 +67,20 @@ abstract class Client {
...
+
+  // Generates and returns incomplete CacaoPayload
+  // To be handled in recaps based authentication.
+  // This enables one-click-auth based authentication to
+  // also enable permissions for notify API.
+  public abstract prepareRegistrationWithRecaps(params: {
+    domain: string;
+    allApps?: boolean;
+  }): Promise<{
+    allApps: boolean;
```
llbartekll, line 80: do we need this boolean in the response?

llbartekll (approved):

jakubuid (approved):

docs/specs/clients/notify/client-sdk-api.md
devceline: It's needed for javascript because of how this will be used by dapps, I imagine the same is true for Swift and Kotlin dapps.

PR #231, opened 2024-05-09
feat(blockchain-api): force balance update for the specific fungible

geekbrother:
This PR updates the Blockchain-API balance endpoint with the ability to force the update of the balance for the specific token using it's contract address to eliminate the balance cache and make UX better.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-server-api.md
+++ b/docs/specs/servers/blockchain/blockchain-server-api.md
@@ -310,13 +310,24 @@ The POST request body should be in JSON format with the following structure:
 
 Used to lookup fungible assets balances
 
-`GET /v1/account/{address}/balance`
+`POST /v1/account/{address}/balance`
 
 #### Path parameters
 
 * `address` - The address for lookup. eg. `0xab16a96D359eC26a11e2C2b3d8f8B8942d5Bfcdb`
 
-#### Query parameters
+#### Request body:
+
+The POST request body should be in JSON format with the following structure:
+
+```typescript
+{
+    "projectId": string,
+    "currency": string,
+    "chain": (optional) string,
+    "forceUpdate": (optional) string[]
+}
+```
 
 * `projectId` - The project identifier.
 * `currency` - Currency to calculate the assets value. e.g. `usd`.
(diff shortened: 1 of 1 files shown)


docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -339,6 +339,7 @@ Used to lookup fungible assets balances
     * `eip155:8508132` - Scroll
     * `eip155:7854577` - Zora
     * `eip155:1313161554` - Aurora
+* `forceUpdate` - (Optional) Token contract address to force the latest balance update on it. `chain` must be also provided for this type of call.
```
enesozturk: Would it be possible to pass array of addresses as `forceUpdate` ? In swaps, we are updating two tokens' balance when swapping, like MATIC -> AVAX, we need to force update both of their balance

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -310,18 +310,29 @@ The POST request body should be in JSON format with the following structure:
...
+#### Request body:
+
+The POST request body should be in JSON format with the following structure:
+
+```typescript
+{
+    "projectId": string,
+    "currency": string,
+    "chain": (optional) string,
+    "forceUpdate": (optional) string[]
```
chris13524: Use actual TypeScript here, is it

```suggestion
    "chain"?: string,
    "forceUpdate"?: string[]
```

or

```suggestion
    "chain": string | null,
    "forceUpdate": string[] | null
```

or both?

docs/specs/servers/blockchain/blockchain-server-api.md
```diff
@@ -339,6 +350,7 @@ Used to lookup fungible assets balances
     * `eip155:8508132` - Scroll
     * `eip155:7854577` - Zora
     * `eip155:1313161554` - Aurora
+* `forceUpdate` - (Optional) List of token contract addresses to force the latest balance update on. `chain` must be also provided for this type of call.
```
enesozturk: It's not clear which type this field is IMO. Is it a string of addresses separated with commas? Or a string array? In which format addresses are (CAIP or contract address)? As Chris mentioned, it's be nice to have TS (or TS like) approach here.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Thanks for pointing out this! It's changed.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: I've updated the spec and changed back to the GET because we need backward compatibility. Also, I've added the description that the expected format is comma separated list of contract addresses in CAIP-10 format.

docs/specs/servers/blockchain/blockchain-server-api.md
geekbrother: Updated to pass a list of comma separated addresses.

enesozturk (approved):

PR #239, opened 2024-06-06
Web3Modal API v2 spec 

bkrem:
## Context
- (Supersedes #228 to fix Git LFS on `main` base branch)
See: https://www.notion.so/walletconnect/Web3Modal-API-v2-Proposal-a2fc6323156d49f79d8092f5c39b8cda

### Scope

- Problem 4 & 5 from the proposal are out-of-scope for the spec, since they are implementation details.

### TODO
- [x] Move current (v1) spec into a separate markdown file where we can back-reference it
- [x] Account for the endpoints that were added previously without updating the v1 spec
- [x] Tie up action items from: https://www.notion.so/walletconnect/Spec-Finalization-Call-a4be73ca50a640eab61a398197708f3c
- [ ] Finalize spec

### DIFF
--- a/docs/specs/meta-clients/web3modal/api.md
+++ b/docs/specs/meta-clients/web3modal/api.md
@@ -26,6 +26,17 @@ All API endpoints expect the following mandatory query parameters:
   - Options: `string`
   - Example: `projectId=702e2d45...`
 
+### Optional Query Parameters
+
+All API endpoints support the following optional query parameters:
+
+- `env=`: Environment
+  - Options: `string`
+  - Example: `env=react-native@0.74.1`
+- `os=`: Operating System
+  - Options: `string`
+  - Example: `os=ios@17.4`
+
 ## Headers
 
 All API endpoints expect the following simple (i.e. non-preflight-dependent) headers:


docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,240 @@
...
-  ledger: "54a1aa77-d202-4f8d-0fb2-5d2bb6db0300",
-  walletConnect: "ef1a1fcf-7fe8-4d69-bd6d-fda1345b4400",
-  injected: "07ba87ed-43aa-4adf-4540-9e6a2b9cae00",
-};
+{
+  isAnalyticsEnabled: boolean,
+}
 ```
+
+## Known Image Sizes
```
bkrem, line 256: @skibitsky I remember you mentioning during the review call that you'd need power-of-2-based image dimensions, are there specific ones we should already define here?

Cali93:
It'd be great to get an answer to [this question](https://www.notion.so/Web3Modal-API-v2-Proposal-a2fc6323156d49f79d8092f5c39b8cda?d=3c79ecb0d086408db8508859c62fe8ac&pvs=4).
At the end of the day, you're consuming this API so anything we can do to improve your DX should be tackled here too.

docs/specs/meta-clients/web3modal/api.md
skibitsky: I ran some tests but didn't see any performance improvements for images downloaded in realtime. As long as the images are square we should be fine! 👍

chris13524 (approved):

glitch-txs:
Hello, since we are moving WalletConnectModal to AppKit we need a way to filter wallets that support the WalletConnect protocol.
e.g Rabby wallet is an extension wallet so we won't need it in the response. However those wallets that are extension but also support mobile (WalletConnect) should be in the response. Coinbase would be another example as they don't support WalletConnect.

cc @enesozturk @tomiir

bkrem:
> Hello, since we are moving WalletConnectModal to AppKit we need a way to filter wallets that support the WalletConnect protocol. e.g Rabby wallet is an extension wallet so we won't need it in the response. However those wallets that are extension but also support mobile (WalletConnect) should be in the response. Coinbase would be another example as they don't support WalletConnect.
> 
> cc @enesozturk @tomiir

@glitch-txs but this doesn't require explicit breaking changes to the API I presume? Seems like either a dedicated endpoint or an additional query parameter so wouldn't have to be settled right now for us to move ahead on implementing the v2 spec.

glitch-txs:
A query parameter would be great

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+- A binary image file (Blob)
 
-There are a number of known static asset IDs that can be used to fetch images, e.g. for networks and connectors:
+### `GET /v2/asset-image/connector/{size}/{connector_id}`
+
+Fetches image for connector logos using descriptive names.
+
+#### Route Parameters
+
+- `connector_id`: Descriptive name for the connector (e.g., `coinbase-wallet`)
```
tomiir: are we using well-known names for these e.g. their rdns? else how do we get the list of available connectors? do we need to hardcode it ?

docs/specs/meta-clients/web3modal/api-v1-legacy.md
```diff
@@ -0,0 +1,142 @@
...
+
+All responses are cached for a minimum of 2h.
+
+## Headers
+
+All api requests must contain following headers. See [types](./types.md)
+
+```ts
+interface Headers {
+  "x-project-id": string;
```
chris13524, line 13: Do these "headers" need the `x-` prefix if we are sending in request body now?

docs/specs/meta-clients/web3modal/api-v1-legacy.md
```diff
@@ -0,0 +1,142 @@
...
+  page: string; // eg. 1
+  entries: string; // eg. 100
+  search?: string; // eg. MetaMa...
+  include?: string; // eg. id1,id2,id3
+  exclude?: string; // eg. id1,id2,id3
+  chains?: string; // eg. eip155:1,solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp
+  platform?: "ios" | "android";
+}
+
+interface Response {
```
chris13524, line 40: In practice how will this be used? When will it be called and with what parameters?

I.e. to get the first page of the modal it should only need to return no more than 5 entries, saving on bandwidth

docs/specs/meta-clients/web3modal/api-v1-legacy.md
```diff
@@ -0,0 +1,142 @@
...
+  count: number;
+  data: Array<{
+    id: string;
+    name: string;
+    homepage: string;
+    image_id: string;
```
chris13524, line 46: Have we considered returning the image inline to save on extra round trips and requests?

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+- WalletConnect: `GET /v2/asset-image/connector/56x56/wallet-connect`
+- Injected (Generic Web3 Injected Connectors): `GET /v2/asset-image/connector/56x56/injected`
+
+#### Response
+
+- A binary image file (Blob)
+
+### `GET /v2/asset-image/token/{size}/{token_symbol}`
+
+Fetches image for known tokens using the token's ticker symbol.
```
chris13524, line 208: Isn't it possible for ticker symbols to overlap? Maybe there is a more robust way to do this? E.g. chain ID + token ID? How is swaps doing this right now?

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+Returns a list of iOS-facing wallet data, including the public Explorer ID and iOS deep link schema.
 
-```ts
-interface Request {
-  headers: Headers;
-}
+#### Response
 
-type Response = Blob;
+```typescript
```
chris13524, line 107: How many entries is this going to return? Why can't the count be limited by the requester? Why isn't this part of `/getWallets`?

docs/specs/meta-clients/web3modal/api-v1-legacy.md
```diff
@@ -0,0 +1,142 @@
...
+
+type Response = Blob;
+```
+
+#### Known Static Asset IDs
+
+There are a number of known static asset IDs that can be used to fetch images, e.g. for networks and connectors:
+
+```typescript
+// Maps EIP155 chain IDs -> `asset_id`
```
chris13524, line 92: Why have a map instead of using the CAIP-2 chain ID directly?

chris13524: Nevermind, this was changed in new API

docs/specs/meta-clients/web3modal/api-v1-legacy.md
tomiir: This is the legacy spec we are changing

docs/specs/meta-clients/web3modal/api-v1-legacy.md
tomiir: Same this is legacy

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
-  entries: string; // eg. 100
-  search?: string; // eg. MetaMa...
-  include?: string; // eg. id1,id2,id3
-  exclude?: string; // eg. id1,id2,id3
-  chains?: string; // eg. eip155:1,solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp
-  platform?: "ios" | "android";
+```typescript
+{
+  "page": number, // Required - The page number of results (e.g., `1`)
+  "entries": number, // Required - Number of entries per page (e.g., `100`)
```
chris13524: Suggest offset and limit instead of page and entries. I.e. right now for the All Wallets view we `exclude` the 4 wallets we got in the first request, but it would better to `offset: 4`

docs/specs/meta-clients/web3modal/api-v1-legacy.md
chris13524: Nevermind, this was changed in new API

docs/specs/meta-clients/web3modal/api.md
tomiir: I agree we at least need to scope this by network given ticker might repeat (eg OP ETH vs Mainnet ETH) 

`.../token/{size}/{caip-2}/{tokenAddress}` would be the most exact, `.../token/{size}/{caip-2}/{symbol}` seems more similar to current structure

docs/specs/meta-clients/web3modal/api-v1-legacy.md
chris13524: Nevermind

docs/specs/meta-clients/web3modal/api-v1-legacy.md
chris13524: Nevermind

docs/specs/meta-clients/web3modal/api.md
Cali93: Agree with Rocky here we should use the `rdns`

bkrem: Agreed, RDNS makes a lot more sense. Will update 👍

docs/specs/meta-clients/web3modal/api.md
bkrem: Good points 👍 

One question: will requiring `tokenAddress` for querying cause more lookup requirements on the SDK side (which we want to explicitly avoid with these changes) or can you query `/{caip-2}/{tokenAddress}` a priori @tomiir?

How is this being queried in the clients currently with the existing `/public/getTokenImage/ETH` approach?

docs/specs/meta-clients/web3modal/api.md
bkrem: @tomiir I just noticed that a number of connectors that are currently being mapped [here in PresetsUtil](https://github.com/WalletConnect/web3modal/blob/main/packages/scaffold-utils/src/PresetsUtil.ts#L10) (e.g. Safe, Ledger) don't have `rdns` values set in Cloud and hence the `/getWallets` response.

Wouldn't this then still necessitate manual mappings in the client either way?

docs/specs/meta-clients/web3modal/api.md
tomiir: @Cali93 was just checking for implementations. 

This was originally intended for token images from Onramp activity- The address is not avail on the response from the tx status https://docs.cdp.coinbase.com/onramp/docs/api-reporting/#transaction-status

but checked the API specs and the response from the options returns the contract address already, so we should have that information available if we set it up correctly

https://docs.cdp.coinbase.com/onramp/docs/api-configurations/#example-requestresponse-1

Given we are requesting this from blockchain API, would it make sense to remove this completely and have blockchain API handle image formatting for that endpoint? 

Other integrations that use token images are not using this endpoint afaik 

Cc @svenvoskamp @enesozturk any cases you can recall where this is being fired? couldn't find anything on the code

chris13524: I was mainly considering a portfolio view; looks like the balances endpoint [returns the contract address](https://specs.walletconnect.com/2.0/specs/servers/blockchain/blockchain-api#success-response-body-5).

chris13524: For onramp API, I would propose that Blockchain API injects the contract address, abstracting away potential conflicts. For now this could be a hardcoded lookup table in Blockchain API

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
-    injected:
-      | {
-          namespace: string;
-          injected_id: string;
-        }[]
-      | null;
-  }>;
+```typescript
+{
+  "count": number,
```
chris13524, line 72: Is this the total number of entries or the length of `data`?

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+    }
+  ]
 }
 ```
 
-### GET `/getWalletImage/:wallet_id`
+### `GET /v2/ios-data`
 
-Returns image logo for requested `wallet_id`. See [types](./types.md).
+Returns a list of iOS-facing wallet data, including the public Explorer ID and iOS deep link schema.
```
chris13524:
```suggestion
Returns a list of iOS-facing wallet data, including the public Explorer ID and iOS deep link schema. This allows apps to check exhaustively what wallets are installed.
```

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+    }
+  ]
+}
 ```
 
-### GET `/public/getAssetImage/:asset_id`
+### `GET /v2/android-data`
 
-Returns image for public asset for requested `asset_id`, i.e network logo, token logo or connector logo.
+Returns a list of Android-facing wallet data, including the public Explorer ID and Android App ID.
```
chris13524:
```suggestion
Returns a list of Android-facing wallet data, including the public Explorer ID and Android App ID. This allows apps to check exhaustively what wallets are installed.
```

docs/specs/meta-clients/web3modal/api.md
chris13524: Should the Coinbase example be `coinbaseWalletSDK`? https://github.com/wevm/wagmi/blob/61483838ec238b406676004a95d08ff702d750d9/packages/connectors/src/coinbaseWallet.ts#L90

docs/specs/meta-clients/web3modal/api.md
chris13524: Unrelated to token asset, there is also the consideration of cache-busting which is not addressed. We have long/immutable-cached responses. For `/v2/wallets` we could change the wallet ID when the image is updated. But for others this is more tricky such as network images.

docs/specs/meta-clients/web3modal/api.md
bkrem: Total number of entries is how this currently works in the v1 API

Wikivu:
Cüzdanlar ile uygulama arasveritabanı bağlamı

chris13524 (commented):
Can we specify caching behavior for these endpoints? I propose:

- response: `Cache-Control: public, max-age=43200` (same as current)
- response: `ETag: <hash of content>`
- request: `If-None-Match: <value of ETag header>`

The request will return a 304 Not Modified if the hash matches, saving bandwidth needing to respond with duplicate data, and renewing the Age of the resource.

docs/specs/meta-clients/web3modal/api.md
chris13524: I propose we use [CAIP-19](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs%2Fcaip-19.md) for this endpoint. So it would look something like this for DAI:

```
GET /v2/asset-image/token/64x64/eip155%3A1%2Ferc20%3A0x6b175474e89094c44da98b954eedeac495271d0f`
```

[eip155 namespace for CAIP-19](https://namespaces.chainagnostic.org/eip155/caip19)

This approach has the added benefit of including the token ID which is important for [ERC-1155](https://eips.ethereum.org/EIPS/eip-1155) tokens as well as NFTs. I.e. having multiple assets for a given contract address. ERC-1155 supersedes ERC-20 and ERC-721 in function.

bkrem:
> Can we specify caching behavior for these endpoints? I propose:
> 
> * response: `Cache-Control: public, max-age=43200` (same as current)
> * response: `ETag: <hash of content>`
> * request: `If-None-Match: <value of ETag header>`
> 
> The request will return a 304 Not Modified if the hash matches, saving bandwidth needing to respond with duplicate data, and renewing the Age of the resource.

Yeah would be great to add this as an optimisation. 

Also it's non-breaking so not strictly a v2 issue, we could in theory already do this right now and clients update their fetch behaviour.

@Cali93 any objections to the `ETag` as an optimised query flow?

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+
+### Mandatory Query Parameters
+
+All API endpoints expect the following mandatory query parameters:
+
+- `st=`: SDK type
+  - Options: `w3m | wcm`
+  - Example: `st=w3m`
+- `sf=`: SDK framework
+  - Options: `html | react | vue`
```
Cali93:
```suggestion
  - Options: `html | react | vue | react-native | flutter | swift | kotlin | unity`
```

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+  - Example: `sf=react`
```
Cali93, line 18: It'd be nice to include the platform as a query param, eg iOS or Android for clients like Flutter and React Native. This will allow deeper validation against bundle IDs (iOS) or package names (anroid).

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
-  chains?: string; // eg. eip155:1,solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp
-  platform?: "ios" | "android";
+```typescript
+{
+  "page": number, // Required - The page number of results (e.g., `1`)
+  "entries": number, // Required - Number of entries per page (e.g., `100`)
+  "search": string | null, // Optional search term (e.g., `MetaMask`)
+  "include": string | null, // Optional comma-separated list of wallet IDs to include
+  "exclude": string | null, // Optional comma-separated list of wallet IDs to exclude
+  "chains": string | null, // Optional comma-separated list of chain identifiers (e.g., `eip155:1`)
```
Cali93: We should also document the behaviour to improve DX when they provide these query params.
Eg what's the logical operator used underneath, what happens if you provide `exclude` and `search` for an excluded `walletId`, etc.

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
+        deep_link: string | null,
+        universal_link: string | null
+      }
+      "injected": [
+        {
+          "namespace": string,
+          "injected_id": string
+        }
+      ] | null
+    }
```
Cali93, line 94: `rnds` is missing and `chains` was added more recently.
```suggestion
      ] | null
    }
    "rdns": string | null
    "chains": string[]
```

docs/specs/meta-clients/web3modal/api.md
Cali93: What's the resolution here ? @tomiir

docs/specs/meta-clients/web3modal/api.md
```diff
@@ -1,142 +1,248 @@
...
 
-#### Known Static Asset IDs
+### `GET /v2/wallet-image/{size}/{wallet_id}`
+
+Returns the image logo for the specified wallet ID.
+
+#### Route Parameters
+
+- `wallet_id`: The unique identifier for the wallet
+- `size`: The predefined image size, described as `{width}x{height}` (see [Known Image Sizes](#known-image-sizes))
```
Cali93, line 144: As a follow up we should document the process for generating new image sizes.

One thing to be aware of, is that with the current approach we're forcing the API consumers to use generic sizes and adjust on their end. So on paper, it means less requests since instead of requesting the same image in 2 different sizes, they would simply scale down the generic image.

With the new approach, if AppKit requests the same image in 56x56 and then in 100x100 it'll result in 2 times more requests and potentially a higher bandwidth consumption.

This is based on the assumption that AppKit requests the same logo in 2 different sizes so we can ignore the concern if it's not/won't be the case.
 
cc @chris13524

Cali93 (approved):
Thanks for driving this 🙏 
Really looking forward to the implementation of this and the results that it'll yield.

docs/specs/meta-clients/web3modal/api.md
ignaciosantise: what about unity? @skibitsky

docs/specs/meta-clients/web3modal/api.md
chris13524: > So on paper, it means less requests since instead of requesting the same image in 2 different sizes, they would simply scale down the generic image.

It _is_ fewer requests, however in real conditions usually the smaller image is requested first, and then 5% of the time the larger image is requested (for example picking a wallet from the list). Furthermore, 100x100=10k vs 56x56=3,136 so it's quadratic improvement to use the smaller image. This results in very large improvements to bandwidth with the cost of 1 additional network request. E.g. requesting 15 3k images and 1 10k image is better than requesting 15 10k images (55k vs 150k). Bigger improvement with more images.

Wikivu:
Verilerim github depomda duruyo ama wallet connect nasıl kullanacağımı
bilmiyorum yardımcı olurmusunuz

Wikivu


29 Tem 2024 Pzt, saat 23:09 tarihinde Chris Smith ***@***.***>
şunu yazdı:

docs/specs/meta-clients/web3modal/api.md
Cali93: Thanks for clarifying and clearing this up 🙏

docs/specs/meta-clients/web3modal/api.md
tomiir: Coinbase we use `coinbaseWalletSDK` as Chris mentions, but you are correct @bkrem that Ledger, Safe and other wallets that are not 6963 might not have it :| 

I think there's no way around the id list then, unless there's a known global ID we can use instead :( 
We'd still need to somehow fetch the id, I guess we already have it from requesting all wallets before

docs/specs/meta-clients/web3modal/api.md
tomiir: Nice, didn't know of CAIP19. It seems like the most explicit given we get the contract address from the API @chris13524

docs/specs/meta-clients/web3modal/api.md
ignaciosantise:
```
env=react-native@0.74.1
os=ios@17.4
```

docs/specs/meta-clients/web3modal/api.md
chris13524, line 94: Do we need ALL of this information in all cases?

Can we shorten the names of the fields to save on bandwidth?

Can we supply as a parameter what platform the app is on and only return the Apple, Google, or mobile links accordingly?

Can we request mobile links/etc. when the wallet is clicked instead of in the initial download?

docs/specs/meta-clients/web3modal/api.md
bkrem: > I think there's no way around the id list then, unless there's a known global ID we can use instead

Yeah the issue at hand seems to be that there's no well-known identifier for these connectors.

> We'd still need to somehow fetch the id

They're all defined here already though right? https://github.com/WalletConnect/web3modal/blob/main/packages/scaffold-utils/src/ConstantsUtil.ts

The only immediate improvement we could make here from what I can see is what is already reflected in the spec, i.e. remove a layer of mapping and query the API by these word-based connectorIDs  directly instead of having this mapping to the imageId on the client/appkit side: https://github.com/WalletConnect/web3modal/blob/main/packages/scaffold-utils/src/PresetsUtil.ts#L78

It means that the mapping (e.g. `coinbaseWalletSDK -> imageId`) needs to live in the API but we already effectively have this burden of maintenance from allowlisting these connectors' imageIds so they can be queried: https://github.com/WalletConnect/cloud-app/blob/develop/services/web3modal-api/src/tools/ConstantsTool.ts#L99

docs/specs/meta-clients/web3modal/api.md
bkrem: Nice one on the CAIP-19 suggestion @chris13524 💯 🙏 

One technicality we need to resolve, without explicit URL-encoding the DAI example is:
```
eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f
```

The slash will be interpreted as an additional path segment, and we can't expect clients to do partial URL-encoding, so we either:
* replace the `/` with another unambiguous value
* Treat the values before and after the `/` like real path segments, i.e. `GET /v2/asset-image/token/{chain_id}/{asset_namespace + ":" + asset_reference}`

Personally I'd prefer to just replace with e.g. an underscore (`eip155:1_erc20:0x6b175474e89094c44da98b954eedeac495271d0f`) because using part of the identifier as a path seems odd.

docs/specs/meta-clients/web3modal/api.md
bkrem: Added here as optional: https://github.com/WalletConnect/walletconnect-specs/pull/239/files#diff-6f25b86fe5ecc8d0a32393be4bab33e4f5b6ca64e95c9180dce7cf7ad76e88deR29

Assuming we don't want to make these mandatory for all possible clients but if we feel strongly about it please let me know 👍

docs/specs/meta-clients/web3modal/api.md
chris13524: > we can't expect clients to do partial URL-encoding

@bkrem what do you mean here? URL encoding is just as easy as base64 encoding. For example in JS with a function such as `encodeURIComponent()`:

```javascript
`GET /v2/asset-image/token/64x64/${encodeURIComponent(caip19Id)}`
```

Keeping it in the original CAIP-19 format sounds simpler to me as you can easily pass the value to another CAIP-19-compatible service or use a lookup table w/o needing to recombine various pieces and use custom encoding.

bkrem: > URL encoding is just as easy as base64 encoding. For example in JS with a function such as encodeURIComponent()

Sure I get the approach, but I think it's odd to expect that kind of accounting work from a client. Seems like this would already get awkward when you want to query this endpoint with e.g. `curl`.

docs/specs/meta-clients/web3modal/api-v1-legacy.md
```diff
@@ -0,0 +1,142 @@
...
+const EIP155NetworkImageIds = {
```
skibitsky, line 93: Ronin is missing

PR #241, opened 2024-06-12
feat(names): updating name registration spec to add mainnet by default

geekbrother:
This PR updates the SPEC description according to the [#669](https://github.com/WalletConnect/blockchain-api/pull/669) changes: adding mainnet cointype by default if non-mainnet cointype provided during the registration.


chris13524 (approved):
LGTM

PR #246, opened 2024-07-26
feat(blockchainAPI): bundler interaction API

geekbrother:
This PR adds a bundler interaction API to the Blockchain-API new `/v1/bundler` endpoint. The user can choose the bundler to call and pass the json-rpc operation to the designated bundler.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-bundler-api.md
+++ b/docs/specs/servers/blockchain/blockchain-bundler-api.md
@@ -1,0 +1,51 @@
+# Blockchain API Bundlers interaction
+
+This API provides a bundlers JSON-RPC interaction. 
+The supported `userOperation` interaction schema is according to v0.7 ERC-4337 specification.
+
+## Make a bundler JSON-RPC request
+
+Used to make a JSON-RPC call to the designated bundler.
+
+`POST /v1/bundler?projectId={projectId}&chainId={chainId}&bundler={bundlerName}`
+
+* `projectId` - Required. The project identifier.
+* `chainId` - CAIP-2 chain identifier.
+* `bundlerName` - (Optional) Bundler name to use.
+
+The list of currently supported bundler names: 
+* `pimlico` - [Pimlico bundler](https://docs.pimlico.io/permissionless/reference/bundler-actions/sendUserOperation).
+
+#### Success response body:
+
+```typescript
+{
+    id:  number, // Unique identifier. Default is 1.
+    jsonrpc: string, // JSON-RPC version. Default 2.0
+    method: string, // ERC-4337 bundler operation methods. Please check supported methods below.
+    params: string[], // Array of string variables input for the method.
+}
+```
+
+List of supported `method` values:
+* `sendUserOperation` - Sends a user operation to the given EVM network.
+* `estimateUserOperationGas` - Estimate the gas limits for a User Operation.
+* `getUserOperationByHash` - Return a User Operation based on a User Operation hash.
+* `getUserOperationReceipt` - Return a User Operation receipt based on a User Operation hash.
+* `supportedEntryPoints` - Return the Entry Points supported by the bundler.
+
+#### Success response body:
+
+Successfull response body will contain the following JSON-RPC structure:
+
+```typescript
+{
+    id: number, // Unique identifier passed to the operation. Default is 1.
+    jsonrpc: string, // JSON-RPC version. Default 2.0
+    result: object // Result object from the operation.
+}
+```
+
+#### Response error codes:
+
+* `400 Bad request` - Wrong input variables format.
--- a/docs/specs/servers/blockchain/blockchain-bundler-api.md
+++ b/docs/specs/servers/blockchain/blockchain-bundler-api.md
@@ -40,7 +40,7 @@ Successfull response body will contain the following JSON-RPC structure:
 {
     id: number, // Unique identifier passed to the operation. Default is 1.
     jsonrpc: string, // JSON-RPC version. Default 2.0
-    result: object // Result object from the operation.
+    result: any // Result object from the operation.
 }
 ```
 
--- a/docs/specs/servers/blockchain/blockchain-bundler-api.md
+++ b/docs/specs/servers/blockchain/blockchain-bundler-api.md
@@ -30,6 +30,7 @@ The list of currently supported bundler names:
 List of supported `method` values:
 * `wallet_getCallsStatus` - Returns the status of a call batch that was sent via `wallet_sendCalls`.
 * `wallet_showCallsStatus` - Requests that a wallet shows information about a given call bundle that was sent with `wallet_sendCalls`.
+* `eth_getUserOperationReceipt` - Get the receipt for the UserOperation.
 
 #### Success response body:
 
--- a/docs/specs/servers/blockchain/blockchain-bundler-api.md
+++ b/docs/specs/servers/blockchain/blockchain-bundler-api.md
@@ -28,11 +28,8 @@ The list of currently supported bundler names:
 ```
 
 List of supported `method` values:
-* `sendUserOperation` - Sends a user operation to the given EVM network.
-* `estimateUserOperationGas` - Estimate the gas limits for a User Operation.
-* `getUserOperationByHash` - Return a User Operation based on a User Operation hash.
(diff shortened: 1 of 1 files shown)


docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,51 @@
...
+{
+    id:  number, // Unique identifier. Default is 1.
+    jsonrpc: string, // JSON-RPC version. Default 2.0
+    method: string, // ERC-4337 bundler operation methods. Please check supported methods below.
+    params: string[], // Array of string variables input for the method.
+}
+```
+
+List of supported `method` values:
+* `sendUserOperation` - Sends a user operation to the given EVM network.
```
geekbrother: Since method names differ in different bundlers (not all, but some) we can unify them as I did here (and convert to the bundler-specific at the server-side) or pass the method through without any changes to the designated bundler.

llbartekll: we need support for wallet_getCallsStatus and wallet_showCallsStatus, Im not aware if we need the others.

llbartekll: https://eips.ethereum.org/EIPS/eip-5792

docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,51 @@
...
+The supported `userOperation` interaction schema is according to v0.7 ERC-4337 specification.
+
+## Make a bundler JSON-RPC request
+
+Used to make a JSON-RPC call to the designated bundler.
+
+`POST /v1/bundler?projectId={projectId}&chainId={chainId}&bundler={bundlerName}`
+
+* `projectId` - Required. The project identifier.
+* `chainId` - CAIP-2 chain identifier.
```
llbartekll, line 13: for eip5792 methods, I'd like to be able to use it without chainId, is it possible?

docs/specs/servers/blockchain/blockchain-bundler-api.md
jakubuid: @geekbrother please add `wallet_getCallsStatus` to the supported list

docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,51 @@
...
+* `bundlerName` - (Optional) Bundler name to use.
+
+The list of currently supported bundler names: 
+* `pimlico` - [Pimlico bundler](https://docs.pimlico.io/permissionless/reference/bundler-actions/sendUserOperation).
+
+#### Success response body:
```
jakubuid:
```suggestion
#### Request Body:
```

docs/specs/servers/blockchain/blockchain-bundler-api.md
llbartekll: Ok, let’s use chainId from session request

docs/specs/servers/blockchain/blockchain-bundler-api.md
geekbrother: I believe we also need to add: `wallet_sendCalls`, or how the client sends the batch to the designated bundler? I believe this should be also done by our endpoint? 
cc @llbartekll @jakubuid

jakubuid: @ganchoradkov what's the implementation here? I thought that wallet is responsible for executing an atomic batch on it's own and this endpoint is for dapp to call for a status

geekbrother: Updated the API: we will allow just `wallet_getCallsStatus` and `wallet_showCallsStatus` operations for now.

docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Make a bundler JSON-RPC request
+
+Used to make a JSON-RPC call to the designated bundler.
+
+`POST /v1/bundler?projectId={projectId}&chainId={chainId}&bundler={bundlerName}`
+
+* `projectId` - Required. The project identifier.
+* `chainId` - CAIP-2 chain identifier.
+* `bundlerName` - (Optional) Bundler name to use.
```
ganchoradkov: >`bundlerName` - (Optional)

is there a default bundler that will be used if the param is missing?

docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,48 @@
...
+    id:  number, // Unique identifier. Default is 1.
+    jsonrpc: string, // JSON-RPC version. Default 2.0
+    method: string, // ERC-4337 bundler operation methods. Please check supported methods below.
+    params: string[], // Array of string variables input for the method.
+}
+```
+
+List of supported `method` values:
+* `wallet_getCallsStatus` - Returns the status of a call batch that was sent via `wallet_sendCalls`.
+* `wallet_showCallsStatus` - Requests that a wallet shows information about a given call bundle that was sent with `wallet_sendCalls`.
```
ganchoradkov, line 32: in the current demo PoC, we're fetching `eth_getUserOperationReceipt` are we going to support it?

docs/specs/servers/blockchain/blockchain-bundler-api.md
geekbrother: The default bundler, for now, will be `Pimlico`. We can make this field required if that's a better approach.

docs/specs/servers/blockchain/blockchain-bundler-api.md
geekbrother: Let's support it for a demo. Updated by adding `eth_getUserOperationReceipt`.

geekbrother: I can add this operation also if needed. Let's support it for now. I have added it.

docs/specs/servers/blockchain/blockchain-bundler-api.md
geekbrother: Fixed. Thanks for catching this!

docs/specs/servers/blockchain/blockchain-bundler-api.md
geekbrother: I changed this to be required since it doesn't makes sense to use default.

ganchoradkov (approved):

ganchoradkov (approved):
Awesome, the api is used now in https://github.com/WalletConnect/walletconnect-monorepo/pull/5171 and https://github.com/WalletConnect/web3modal/pull/2666 💯 💯

docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,49 @@
...
+
+#### Success response body:
+
+Successfull response body will contain the following JSON-RPC structure:
+
+```typescript
+{
+    id: number, // Unique identifier passed to the operation. Default is 1.
+    jsonrpc: string, // JSON-RPC version. Default 2.0
+    result: object // Result object from the operation.
```
chris13524: `object` isn't valid TypeScript, use `any` instead:

```suggestion
    result: any // Result object from the operation.
```

docs/specs/servers/blockchain/blockchain-bundler-api.md
```diff
@@ -0,0 +1,49 @@
...
+
+The list of currently supported bundler names: 
+* `pimlico` - [Pimlico bundler](https://docs.pimlico.io/permissionless/reference/bundler-actions/sendUserOperation).
+
+#### Request body:
+
+```typescript
+{
+    id:  number, // Unique identifier. Default is 1.
+    jsonrpc: string, // JSON-RPC version. Default 2.0
```
chris13524, line 24: I don't feel like we should have defaults for these fields. This looks like standard JSON-RPC and our RPC proxy endpoints don't do this

chris13524 (approved):
LGTM

docs/specs/servers/blockchain/blockchain-bundler-api.md
geekbrother: Fixed. Thanks for catching this!

PR #247, opened 2024-07-30
feat(blockchainAPI): adding `v=2` optional query parameter for name lookups

geekbrother:
This PR adds an optional `apiVersion=2` query parameter for name lookups to fix the [404 response issue](https://github.com/WalletConnect/blockchain-api/issues/700) and support backward compatibility.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-api.md
+++ b/docs/specs/servers/blockchain/blockchain-api.md
@@ -45,6 +45,10 @@ Used to lookup address for the name.
     * Case insensitive alphabetic characters (a-z), numeric characters (0-9), the minus sign (-) and the period (.).
     * Max 255 characters length.
 
+#### Query parameters:
+
+* `v` - (Optional) API version specification. `1` by default. Can be `2` to force using the updated API response.
+
 #### Success response body:
 
 ```typescript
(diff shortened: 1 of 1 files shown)


tomiir (approved):

docs/specs/servers/blockchain/blockchain-api.md
```diff
@@ -45,6 +45,10 @@ Used to lookup address for the name.
     * Case insensitive alphabetic characters (a-z), numeric characters (0-9), the minus sign (-) and the period (.).
     * Max 255 characters length.
 
+#### Query parameters:
```
nopestack, line 48: Would it be too difficult to support v1 and v2 prefixes for the APIs instead? e.g. `/v1/my-endpoint?query=value` and `/v2/endpoint?my_query=value`

nopestack: Also if ☝️ isn't applicable in this context, consider naming the parameter `version` or `api_version` instead as it's more explicit about what it represents

tomiir: Agree not using exactly the sane name as the route path would be better

geekbrother: Agree! Makes sense to change it to the `api_version`. I'll make this change now.

cc @nopestack We had the [same conversation in this thread recently](https://walletconnect.slack.com/archives/C03SCF66K2T/p1722097238444779) and stick to `v=2` `(api_version=2)` and omit creating new `/v2` endpoint, because the changes are tiny and this is actually a hotfix neither the schema change.

nopestack: ok I read the discussion, I don't entirely agree with the rationale of going with a query parameter but folks agree that this change is small enough not to be considered a `/v2` 👍

nopestack (approved):
Do consider the parameter rename to something more explicit before merging though

docs/specs/servers/blockchain/blockchain-api.md
chris13524: If I was being pure I would do a `/v2` endpoint, I don't think it has anything to do with schema change vs other type of change. But pragmatically it's a lot easier to implement via a query param.

chris13524 (approved):

geekbrother:
The parameter name was changed to `apiVersion`.

PR #248, opened 2024-08-07
feat(relay): request in connection URL

chris13524:


docs/specs/servers/relay/relay-server-api.md
```diff
@@ -2,12 +2,16 @@
 
 ## WebSocket
 
-WebSocket supports JSON-RPC methods. For more details, please visit [Relay RPC docs](./relay-server-rpc.md).
+The `/` endpoint is used to establish WebSocket connections. Messages sent are JSON-RPC requests and responses are JSON-RPC responses. See [supported JSON-RPC methods](./relay-server-rpc.md) for valid JSON-RPC payloads.
```
chris13524, line 5: That part's not true

```suggestion
The `/` endpoint is used to establish WebSocket connections. See [supported JSON-RPC methods](./relay-server-rpc.md) for valid JSON-RPC payloads.
```

docs/specs/servers/relay/relay-server-api.md
```diff
@@ -2,12 +2,16 @@
...
 
-## HTTP
+### Query parameters
 
-### JSON-RPC
+- `auth` (required) - A JWT for the client
+- `projectId` (required) - The project ID
+- `request` - A base64-encoded [JSON-RPC method](./relay-server-rpc.md) to be executed in-parallel with the connection being established. The JSON-RPC response is sent over the WebSocket connection to the client.
```
chris13524, line 11:
```suggestion
- `request` - A base64-encoded [JSON-RPC request](./relay-server-rpc.md) to be executed in-parallel with the connection being established. The JSON-RPC response is sent over the WebSocket connection to the client.
```

docs/specs/servers/relay/relay-server-api.md
```diff
@@ -2,12 +2,16 @@
...
+- `ua` - A user agent string
+- `useOnCloseEvent` - unused?
```
jakubuid, line 13: What's that?

jakubuid: Let's add an example of such connection URL

heilhead: It's a boolean flag that JS SDK used back in the day (maybe still using?) for the relay to send an error back as a JSON RPC error frame before closing the connection, if an error happenned during the client validation. The problem was that the JS websocket client can't distinguish between different rejection/close codes.

PR #249, opened 2024-08-09
feat(blockchainAPI): adding `namespace` optional parameter for balance and history requests

geekbrother:
This PR adds an optional `namespace` request parameter to the address balance and transactions history endpoint.
Since we will support Solana for the balance and transactions soon we need to distinguish namespaces for the query address and support backward compatibility.

### DIFF
--- a/docs/specs/servers/blockchain/blockchain-api.md
+++ b/docs/specs/servers/blockchain/blockchain-api.md
@@ -343,6 +343,7 @@ Used to lookup fungible assets balances
 {
     "projectId": string,
     "currency": string,
+    "namespace"?: string,
     "chain"?: string,
     "forceUpdate"?: string
 }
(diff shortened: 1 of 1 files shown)


docs/specs/servers/blockchain/blockchain-api.md
```diff
@@ -394,24 +398,29 @@ Used to lookup fungible assets balances
 
 Used to lookup transactions list for an address.
 
-`GET /v1/account/{address}/history`
+`GET /v1/account/{address}/history?projectId={projectId}`
 
 #### Path parameters
 
 * `address` - The address for lookup. eg. `0xab16a96D359eC26a11e2C2b3d8f8B8942d5Bfcdb`
+* `projectId` - The project identifier.
```
geekbrother, line 406: Adding the `projectId` here just because it was missed.

chris13524 (commented):
Wouldn't it be better to make the `address` path param a CAIP-10?

geekbrother:
> Wouldn't it be better to make the `address` path param a CAIP-10?

We are returning results for all popular tokens in chain IDs, by default. 
In case of CAIP-10 we should provide some chain ID. The closest is CAIP-19, but it still lacks what we are looking for, like `namespace:address` without specifying chain ID.

chris13524 (approved):

geekbrother:
Closing this PR because we've decided to use just `chainId`. The [full Slack context](https://walletconnect.slack.com/archives/C03SCF66K2T/p1724246068730469?thread_ts=1722872189.697099&cid=C03SCF66K2T).

